PR #60, opened 2022-04-05
add label handle method for PipelineHandler

tianweidut:
## Description
user can define label handler which can make compare with result easily. default label will output as str type.

In our MNIST example, we define `handle_label` method,
```python
def handle_label(self, label, batch_size, **kw):
     return [int(l) for l in label]
```
image

## Modules
- [x] Python-SDK

## Checklist
- [ ] add unit test
- [ ] add necessary doc


goldenxinxing (approved):
lgtm

PR #147, opened 2022-04-27
use _manifest runtime as the first class runtime for python

tianweidut:
## Modules
- [x] Python-SDK
- [x] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


anda-ren (approved):
lgtm

Issue #345, opened 2022-05-17
eval run error in docker 19.03 environment


tianweidut:
**Describe the bug**
![image](https://user-images.githubusercontent.com/590748/168710133-75b49423-6590-4b9a-a117-a1394f348bda.png)

**To Reproduce**
in docker client version 19.03.

**Expected behavior**
pull image, run container.
![image](https://user-images.githubusercontent.com/590748/168710223-dfa7968b-a5eb-4563-9140-b34791c24d9c.png)
use docker pull cmd in advance.

**Environment**
![image](https://user-images.githubusercontent.com/590748/168710183-28a1359d-f200-4999-bafe-21ac9f3ef3c8.png)

PR #346, opened 2022-05-17
fix: use docker pull cmd for prepare phase

tianweidut:
## Description
- only the newer version(>= 20.10) support `docker run --pull=always` args, so we should use `docker pull` cmd to pull image.
- issue: #345 

## Modules
- [x] Client

## Checklist
- [ ] add unit test
- [ ] add necessary doc


goldenxinxing (approved):
lgtm

PR #506, opened 2022-06-07
Runtime push force & head dataset api

dreamlandliu:
## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


tianweidut:
👏

server/controller/src/main/java/ai/starwhale/mlops/domain/swds/SWDatasetService.java
```diff
@@ -286,4 +287,16 @@ private List<SWDatasetInfoVO> swDatasetInfoOfDs(SWDatasetEntity ds) {
         return swDatasetVersionEntities.parallelStream()
             .map(entity -> toSWDatasetInfoVO(ds, entity)).collect(Collectors.toList());
     }
+
+    public String query(UploadRequest uploadRequest) {
+        SWDatasetEntity entity = swdsMapper.findByNameForUpdate(uploadRequest.name());
```
anda-ren, line 292: we do not lock the row when in pure query process, so I think `findByName` is sufficent

anda-ren (approved):
lgtm, except that the lock is not needed

PR #615, opened 2022-06-21
RestAPI & Service & Mapper & Objects for Evaluation

dreamlandliu:
## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


server/controller/src/main/resources/db/migration/V00003__Add_view_config.sql
```diff
@@ -0,0 +1,15 @@
+-- ----------------------------
```
anda-ren, line 1: Incremental sql scripts are used between releases. In scope of one release, we should add the script to the only one script file that belongs to this release

anda-ren (changes requested):
some small issues

anda-ren (approved):
lgtm

Issue #611, opened 2022-06-21
more user friendly to serialize tensor & numpy array


anda-ren:
torch.tensor numpy.array can't be serialized to json directlly, both of them should be transfered to multi-dimision list before serialized to json. 

**Current Behavior**
All results of `PipelineHandler.ppl` are serialized to JSON. If user return `tensor` or `numpy` object in `ppl` method, error will be raised.
If user transfers `tesnor` or `numpy` to  multi-dimision list , precision of decimal number will be lost in some degree.

Here is the example which illustrates this issue :https://github.com/star-whale/starwhale/blob/main/example/PennFudanPed/code/ppl.py
```python
 @torch.no_grad()
    def ppl(self, data, batch_size, **kw):
        model = self._load_model(self.device)
        image = Image.open(io.BytesIO(data))
        _image = F.to_tensor(image)
        outputs = model([_image.to(self.device)])
        cpu_device = torch.device("cpu")
        # [{'boxes':tensor[[],[]]},'labels':tensor[[],[]],'masks':tensor[[[]]]}]
        outputs = [{k: v.to(cpu_device) for k, v in t.items()} for t in outputs]
        for t in outputs:
            #convert tensor to multi-dimission list
            self.tensor_dict_to_list_dict(t) 
            t['height'] = _image.shape[-2]
            t['width'] = _image.shape[-1]
        return outputs, None
```

```python
def cmp(self, _data_loader):
        _result, _label = [], []
        for _data in _data_loader:
           #convert list to tensor
            _label.extend([self.list_dict_to_tensor_list(l, True) for l in _data["label"]]) 
            _result.extend([self.list_dict_to_tensor_list(r, False) for r in _data["result"]]) 
        ds = zip(_result, _label)
        coco_ds = coco_utils.convert_to_coco_api(ds)
        coco_evaluator = coco_eval.CocoEvaluator(coco_ds,  ["bbox", "segm"])
        for outputs, targets in zip(_result, _label):
            res = {targets["image_id"].item(): outputs}
            coco_evaluator.update(res)

        # gather the stats from all processes
        coco_evaluator.synchronize_between_processes()

        # accumulate predictions from all images
        coco_evaluator.accumulate()
        coco_evaluator.summarize()

        return [{iou_type: coco_eval.stats.tolist() for iou_type, coco_eval in coco_evaluator.coco_eval.items()}]
```

**Proposed Behavior**
- there should be no restriction of what type user returned in `ppl` method. 
[...]

PR #631, opened 2022-06-23
feat(sdk): support serialize and deserialize in PipelineHandler

jialeicui:
## Description
support serialize and deserialize in PipelineHandler
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [x] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/api/_impl/model.py
+++ b/client/starwhale/api/_impl/model.py
@@ -4,6 +4,7 @@ import io
 import os
 import sys
 import json
+import base64
 import typing as t
 import logging
 from abc import ABCMeta, abstractmethod
@@ -11,6 +12,7 @@ from types import TracebackType
 from pathlib import Path
 from functools import wraps
 
+import dill
 import loguru
 import jsonlines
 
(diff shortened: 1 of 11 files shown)


client/starwhale/api/_impl/model.py
```diff
@@ -288,36 +304,36 @@ def _starwhale_internal_run_ppl(self) -> None:
...
+        *args,
     ) -> None:
         self._status_writer.write(
             {
                 "time": now_str(),  # type: ignore
                 "status": exception is None,
                 "exception": str(exception),
                 "index": data.idx,
-                "output_size": len(pred),
+                "output_size": len(args[0]),
```
tianweidut, line 322: if `args` is empty, `args[0]` will raise IndexError ?

tianweidut (commented):
Other LGTM

tianweidut:
#611

client/starwhale/api/_impl/model.py
jialeicui: can we assume that `ppl`  function will always return something for `cmp`?

tianweidut: > can we assume that `ppl` function will always return something for `cmp`?

yes.

anda-ren (commented):
LGTM

anda-ren (approved):
lgtm

tianweidut (approved):
LGTM

PR #690, opened 2022-06-30
fix(client): make targets phony

jialeicui:
## Description

* make virtual targets PHONY
* `all-check` can not autocomplete without phony under zsh

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


tianweidut (approved):
LGTM

PR #769, opened 2022-07-20
ci: chang e2e test running rate

anda-ren:
- run at 2:34 am and 5:34 am at UTC+8
- use github filter instead of market filter

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [x] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


tianweidut:
Does `getsentry/paths-filter@v2` have some problems ?  @anda-ren

tianweidut:
Use another explicated title?

.github/workflows/docs-ignore.yaml
```diff
@@ -0,0 +1,19 @@
+name: Render and Publish docs
+
+on:
+  push:
+    branches:
+      - main
+    paths-ignore:
```
tianweidut, line 7: In the main branch, we should run all the ci.

Issue #741, opened 2022-07-17
Starwhale Runtime 2.0


tianweidut:
**Current Behavior**

```yaml
mode: venv
name: pytorch
pip_req: requirements.txt
python_version: '3.9'
starwhale_version: ''
```

**Proposed Behavior**

```yaml
version: 2.0

name: pytorch  #required: runtime name
mode: venv     #optional, default venv

environment:
  arch: noarch     #optional, x86/aarch64/noarch
  os: linux            #optional 
  python: 3.8.13  #optional, python version
  cuda: xxx          #optional, cuda version

dependencies:          # optional
  - requirements.txt   

  - conda.yaml         
  - conda:
     - package==version   # conda python version
  - pip:
     - package==version   # python package version
  - wheels:
     - xxx.whl            # wheel package uri(local or remote)
  - files:
     - name: xxx          
       src: xxx          
       dest: yyy          
       pre: cmd           # hook: pre command
       post: cmd          # hook: post command
       
 hooks:          # optional
   pre: xxx      # restore bash
   post: yyy     # restore bash
   
 configs:        # optional
   docker:       
      registry: ghcr.io
      image: star-whale/starwhale
   conda:
      channels:
        - conda-forge  # conda channels
   pip:
      index_url: xx
      external_url: xx
      trusted_hosts: xx
```

**Subtasks**
- [x] https://github.com/star-whale/starwhale/issues/742
- [x] https://github.com/star-whale/starwhale/issues/744
- [x] https://github.com/star-whale/starwhale/issues/745
- [x] https://github.com/star-whale/starwhale/issues/747
- [x] https://github.com/star-whale/starwhale/issues/749
- [x] Tune activate
    - [x] https://github.com/star-whale/starwhale/issues/572
    - [x] https://github.com/star-whale/starwhale/issues/752
- [x] https://github.com/star-whale/starwhale/issues/751   
- [x] runtime dockerize
   - [x] https://github.com/star-whale/starwhale/issues/750
   - [x] #522

PR #788, opened 2022-07-21
feat(runtime): support conda install to use configs.channels field in the runtime.yaml

tianweidut:
## Description
- #741 

## Modules
- [x] Client

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/utils/venv.py
+++ b/client/starwhale/utils/venv.py
@@ -19,6 +19,7 @@ from starwhale.consts import (
     ENV_CONDA_PREFIX,
     SW_PYPI_PKG_NAME,
     WHEEL_FILE_EXTENSION,
+    DEFAULT_CONDA_CHANNEL,
     DEFAULT_PYTHON_VERSION,
 )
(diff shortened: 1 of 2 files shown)


client/starwhale/utils/venv.py
```diff
@@ -707,6 +722,7 @@ def _do_restore_venv(
     _python_version: str,
     _wheels: t.Optional[t.List[str]],
     _deps: t.Optional[t.Dict[str, t.Union[t.List[str], str]]],
+    _configs: t.Optional[t.Dict[str, t.Dict[str, t.Union[str, t.List[str]]]]],
```
jialeicui, line 725: consider type alias? 
https://docs.python.org/3/library/typing.html#type-aliases

jialeicui:
LGTM

client/starwhale/utils/venv.py
tianweidut: ok

tianweidut: updated

jialeicui (approved):

Issue #765, opened 2022-07-18
tune `runtime build` with `--lock/--disable-lock`


tianweidut:
**Current Behavior**
Generate lock files automatically, but users cannot inspect the file details.

**Proposed Behavior**
- `swcli runtime build --lock` command will dump the lock file and compare contents with [locked file](https://github.com/star-whale/starwhale/issues/764), give users some advice.
- `--lock` as the default args.
- Users can config `--disable-lock` to disable requirements lock.

tianweidut:
https://github.com/star-whale/starwhale/pull/807

PR #807, opened 2022-07-25
feat(runtime): add `enable-lock` for swcli build and refactor runtime build/restore process with runtime.yaml 2.0 definition

tianweidut:
## Description
- `swcli runtime build` does not depend on the current shell environment.
- `swcli runtime build` process is explicit in that it only uses `runtime.yaml` to build swrt. 
- Simplify to generate venv.tar in local_gen mode.
- related issues: 
  - https://github.com/star-whale/starwhale/issues/741
  - https://github.com/star-whale/starwhale/issues/765

## Feature Show
- no shell environment dependency
image
- venv-local-gen/restore

image
- conda-local-gen/restore
image
image
- conda build/restore
image
image

## Modules
- [x] Client

## Checklist
- [x] add unit test
- [ ] add necessary doc


client/starwhale/utils/venv_pack.py
```diff
@@ -0,0 +1,31 @@
...
+    from .venv import check_valid_venv_prefix
+
+    if not check_valid_venv_prefix(prefix):
+        raise FormatError(f"venv prefix: {prefix}")
+
+    # TODO: check editable packages
+    temp_dir = tempfile.mkdtemp(prefix="starwhale-venv-", dir=os.path.dirname(output))
+
+    try:
+        copy_fs(str(prefix), temp_dir)
```
jialeicui, line 26: Could you please add some comments for the reason copying files to a temp dir

client/starwhale/utils/venv_pack.py
```diff
@@ -0,0 +1,31 @@
...
+
+        with tarfile.open(output, "w:gz") as tar:
+            tar.add(str(temp_dir), arcname="")
```
jialeicui, line 29: Will tar to temp file and make an atomic mv be better?

tianweidut: ok

client/starwhale/utils/venv_pack.py
tianweidut: Previous points:
venv dir --> copy to another unused dir -> tar the unused dir.
Making tar.gz will cost a lot of time if it works on a user's venv dir that may be changed. tar.gz will be broken.

But when I test the according scene， it works well.  So the copy_fs will be removed.

jialeicui (approved):

PR #830, opened 2022-07-28
add the first edition of starwhale sdk

xuchuan:
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [X] Python-SDK
- [ ] Others

## Checklist
- [X] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/setup.py
+++ b/client/setup.py
@@ -24,6 +24,7 @@ install_requires = [
     "scikit-learn>=0.20.0",
     "dill==0.3.5.1",
     "packaging>=21.3",
+    "pyarrow>=8.0.0",
 ]
 
 
@@ -44,7 +45,7 @@ setup(
     author_email="developer@starwhale.ai",
     version=_format_version(),
     description="An MLOps Platform for Model Evaluation",
-    long_description=open("../README.md").read(),
+    long_description=open("../README.md", encoding="utf-8").read(),
     long_description_content_type="text/markdown",
     keywords="MLOps, AI, Starwhale, Model Evaluation",
     url="https://github.com/star-whale/starwhale",
--- a/client/starwhale/api/_impl/data_store.py
+++ b/client/starwhale/api/_impl/data_store.py
@@ -1,0 +1,808 @@
+import os
+import re
+import sys
+import json
+import atexit
+import pathlib
+import threading
+from typing import Any, Set, cast, Dict, List, Tuple, Iterator, Optional
+
+import numpy as np
+import pyarrow as pa  # type: ignore
+import pyarrow.parquet as pq  # type: ignore
+from typing_extensions import Protocol
+
+
+class Type:
+    def __init__(
+        self, name: str, pa_type: pa.DataType, nbits: int, default_value: Any
+    ) -> None:
+        self.name = name
+        self.pa_type = pa_type
+        self.nbits = nbits
+        self.default_value = default_value
+
+    def serialize(self, value: Any) -> Any:
+        return value
+
+    def deserialize(self, value: Any) -> Any:
+        return value
+
+    def __str__(self) -> str:
+        if self.name == "int" or self.name == "float":
+            return f"{self.name}{self.nbits}"
+        else:
+            return self.name
+
+    __repr__ = __str__
+
+
+class LinkType(Type):
+    def __init__(self) -> None:
+        super().__init__("link", pa.string(), 32, "")
+
+    def serialize(self, value: Any) -> Any:
+        if value is None:
+            return None
+        assert isinstance(value, Link)
+        return str(value)
+
+    def deserialize(self, value: Any) -> Any:
+        if value is None:
+            return None
+        d = json.loads(value)
+        return Link(
+            d.get("uri", None), d.get("display_text", None), d.get("mime_type", None)
+        )
+
+
+class Link:
+    def __init__(
+        self,
+        uri: str,
+        display_text: Optional[str] = None,
+        mime_type: Optional[str] = None,
+    ) -> None:
+        self.uri = uri
+        self.display_text = display_text
+        self.mime_type = mime_type
+
+    def __str__(self) -> str:
+        return json.dumps(
+            {
+                "uri": self.uri,
+                "display_text": self.display_text,
+                "mime_type": self.mime_type,
+            }
+        )
+
+    def __eq__(self, other: Any) -> bool:
+        return (
+            isinstance(other, Link)
+            and self.uri == other.uri
+            and self.display_text == other.display_text
+            and self.mime_type == other.mime_type
+        )
+
+
+NONE = Type("none", None, 1, None)
+INT8 = Type("int", pa.int8(), 8, 0)
+INT16 = Type("int", pa.int16(), 16, 0)
+INT32 = Type("int", pa.int32(), 32, 0)
+INT64 = Type("int", pa.int64(), 64, 0)
+FLOAT16 = Type("float", pa.float16(), 16, 0.0)
+FLOAT32 = Type("float", pa.float32(), 32, 0.0)
+FLOAT64 = Type("float", pa.float64(), 64, 0.0)
+BOOL = Type("bool", pa.bool_(), 1, 0)
+STRING = Type("str", pa.string(), 32, "")
+BYTES = Type("bytes", pa.binary(), 32, b"")
+LINK = LinkType()
+
+_TYPE_DICT: Dict[Any, Type] = {
+    type(None): NONE,
+    np.byte: INT8,
+    np.int8: INT8,
+    np.int16: INT16,
+    np.int32: INT32,
+    np.int64: INT64,
+    int: INT64,
+    np.float16: FLOAT16,
+    np.float32: FLOAT32,
+    np.float_: FLOAT64,
+    float: FLOAT64,
+    np.bool_: BOOL,
+    bool: BOOL,
+    str: STRING,
+    bytes: BYTES,
+    Link: LINK,
+}
+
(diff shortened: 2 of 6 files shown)


sdk/python/starwhale/data_store.py
```diff
@@ -0,0 +1,703 @@
...
+
+    def __repr__(self) -> str:
+        return self.__str__()
+
+    def __eq__(self, other) -> bool:
+        return isinstance(other, TableSchema) and self.key_column == other.key_column and self.columns == other.columns
+
+
+def _get_table_path(root_path: str, table_name: str) -> str:
+    return os.path.join(root_path, *table_name.split("/"))
```
tianweidut, line 150: if the `table_name` is a path split string, `os.path.join` can handle it automatically.

```
In [1]: import os

In [2]: os.path.join("a", "b/c/d/e")
Out[2]: 'a/b/c/d/e'
```

tianweidut:
@xuchuan  move the SDK code into [client/starwhale/api/_impl](https://github.com/star-whale/starwhale/tree/main/client/starwhale/api/_impl) , and move the test code into client/starwhale/tests ?

Benefits:
1. client dir is a root python package dir. The code will be packaged into the PyPI pkg.
2. Python code linters and tests ci will work.
3. Starwhale other python code and user code can import the `data_store` module.

sdk/python/starwhale/data_store.py
```diff
@@ -0,0 +1,703 @@
+from typing import Any, Tuple, Optional, Iterator
+import atexit
+import json
+import os
+import re
+import sys
+import threading
+
```
tianweidut, line 8: use [black](https://github.com/star-whale/starwhale/blob/main/client/Makefile#L37) to format the python code? The useless blank line will be removed

sdk/python/starwhale/data_store.py
```diff
@@ -0,0 +1,703 @@
...
+
+import numpy as np
+import pyarrow as pa
```
tianweidut, line 11: add `pyarrow` req into [setup.py](https://github.com/star-whale/starwhale/blob/main/client/setup.py#L5) ?

sdk/python/starwhale/data_store.py
```diff
@@ -0,0 +1,703 @@
...
+
+
+class Link:
+    def __init__(self, uri: str, display_text: Optional[str] = None, mime_type: Optional[str] = None):
+        self.uri = uri
+        self.display_text = display_text
+        self.mime_type = mime_type
+
+    def __str__(self) -> str:
+        return json.dumps({"uri": self.uri, "display_text": self.display_text, "mime_type": self.mime_type})
```
tianweidut, line 62: add `__repr__ = __str__` ?

sdk/python/starwhale/data_store.py
```diff
@@ -0,0 +1,703 @@
...
+
+    def __eq__(self, other) -> bool:
```
tianweidut, line 64: add a type for `other` argument ?

sdk/python/starwhale/data_store.py
```diff
@@ -0,0 +1,703 @@
...
+        return isinstance(other, ColumnSchema) and self.name == other.name and self.type is other.type
+
+
+class TableSchema:
+    def __init__(self, key_column: str, columns: list[ColumnSchema]):
+        self.key_column = key_column
+        self.columns = {col.name: col for col in columns}
+
+    def copy(self) -> "TableSchema":
```
tianweidut, line 126: when write `from __future__ import annotations` at the first line,  `"TableSchema"` can be replaced with `TableSchema`. It is friendly to IDE.

sdk/python/starwhale/data_store.py
```diff
@@ -0,0 +1,703 @@
...
+    try:
+        if name.endswith(".parquet"):
+            if name.startswith("base-"):
+                return "base", int(name[5:-8])
+            elif name.startswith("patch-"):
+                return "patch", int(name[6:-8])
+    except ValueError:
+        # ignore invalid filename
+        pass
+    return None, None
```
tianweidut, line 163: type error?  the expected return type is `Tuple[str, int]` ?

sdk/python/starwhale/data_store.py
```diff
@@ -0,0 +1,703 @@
...
+    with pq.ParquetWriter(filename, table.schema) as writer:
+        writer.write_table(table)
+
+
+def _scan_parquet_file(
+    path: str, columns: Optional[dict[str, str]] = None, start: Optional[Any] = None, end: Optional[Any] = None
+) -> Iterator[dict]:
+    f = pq.ParquetFile(path)
+    schema_arrow = f.schema_arrow
+    if columns is None:
```
tianweidut, line 176: does the empty colums dict should be set the `{name: name for name in schema_arrow.names if name != "-" and not name.startswith("~")}` ?

```
if columns is None:
    columns = {name: name for name in schema_arrow.names if name != "-" and not name.startswith("~")}
```
-->
```
columns = columns or {name: name for name in schema_arrow.names if name != "-" and not name.startswith("~")}
```

sdk/python/starwhale/wrapper.py
```diff
@@ -0,0 +1,94 @@
...
+        else:
+            for k, v in kwargs.items():
+                record[k.lower()] = v
+        self._log(self._summary_table_name, record)
+
+    def get_results(self) -> Iterator[dict[str, Any]]:
+        return self._data_store.scan_tables([(self._results_table_name, "result", False)])
+
+
+class Dataset(Logger):
```
anda-ren, line 71: shall we introduce another lower level dataset which is independent of project. Mybe the one here could be called `SWDataset`, while that lower level one could be called `Dataset`

sdk/python/starwhale/data_store.py
xuchuan: it does not work on Windows, which will be "a\\b/c/d/e"

sdk/python/starwhale/data_store.py
xuchuan: it is on purpose. I use blanks to separate Python built-in libraries and third-party libraries

sdk/python/starwhale/data_store.py
xuchuan: I prefer the old way so that it can work in a low-version Python environment. IDE can handle this old-fashion typing annotation well.

sdk/python/starwhale/data_store.py
xuchuan: I guess it is ok to return None for int?

sdk/python/starwhale/wrapper.py
xuchuan: I prefer Dataset because it is recommended to use it like this:
```
import starwhale as sw
ds = sw.Dataset()
```
`sw.SWDataset` looks really weird

anda-ren: I didn't take account the use case. You are right!

sdk/python/starwhale/data_store.py
```diff
@@ -0,0 +1,703 @@
...
+
+    def __str__(self) -> str:
+        return json.dumps(
+            {
+                "key": self.key_column,
+                "columns": [{"name": col.name, "type": str(col.type)} for col in self.columns.values()],
+            }
+        )
+
+    def __repr__(self) -> str:
```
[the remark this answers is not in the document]
xuchuan: done

xuchuan:
> @xuchuan move the SDK code into [client/starwhale/api/_impl](https://github.com/star-whale/starwhale/tree/main/client/starwhale/api/_impl) , and move the test code into client/starwhale/tests ?
> 
> Benefits:
> 
> 1. client dir is a root python package dir. The code will be packaged into the PyPI pkg.
> 2. Python code linters and tests ci will work.
> 3. Starwhale other python code and user code can import the `data_store` module.

done

sdk/python/starwhale/data_store.py
xuchuan: done

sdk/python/starwhale/data_store.py
xuchuan: done

sdk/python/starwhale/data_store.py
xuchuan: done

sdk/python/starwhale/data_store.py
xuchuan: columns should not be empty here. As it is an optional parameter, users should provide a meaningful value when they specify it explicitly. So I would like to return an empty list for empty columns rather than treat it as the same as None.

sdk/python/starwhale/data_store.py
tianweidut: The `pathlib` is maybe  the better choice.

- in linux

```python
In [1]: import platform
In [2]: platform.system()
Out[2]: 'Linux'
In [3]: from pathlib import Path
In [4]: Path("/tmp") / "b/c/d"
Out[4]: PosixPath('/tmp/b/c/d')
```

- in windows

```python
>>> import platform
>>> platform.system()
'Windows'
>>> from pathlib import Path
>>> Path("/tmp") / "b/c/d"
WindowsPath('/tmp/b/c/d')
>>> Path("/tmp") / "b\c\d"
WindowsPath('/tmp/b/c/d')
```
cc @xuchuan

xuchuan: done

sdk/python/starwhale/data_store.py
xuchuan: Done

goldenxinxing (approved):
LGTM

Issue #752, opened 2022-07-17
tune the user experience of Starwhale rutime activating


tianweidut:
**Current Behavior**
Effortless console commands output.

**Proposed Behavior**
- Activate Scope:
   - Shell Activate
   - Swcli command Activate
   - Docker runtime container
- Subtasks: 
   - [x] activate from runtime uri: https://github.com/star-whale/starwhale/pull/850
   - [x] Add `--runtime` argument for some `swcli` commands, and the python environment will automatically activate. (https://github.com/star-whale/starwhale/pull/845)
   - [x] Activate python runtime when users start/exec Starwhale Runtime docker container that the image is generated by `swcli runtime dockerize` command.  (https://github.com/star-whale/starwhale/pull/840  https://github.com/star-whale/starwhale/pull/848)

PR #848, opened 2022-08-02
enhance(runtime): add runtime-restore args for RuntimeProcess and auto extract for runtime restore

tianweidut:
## Description
- add extract for standalone runtime restore
- add runtime-restore args for RuntimeProcess class and job create cmd
- related: https://github.com/star-whale/starwhale/issues/752

## Modules
- [x] Client

## Checklist
- [ ] add unit test
- [ ] add necessary doc


client/starwhale/utils/venv.py
```diff
@@ -440,22 +447,25 @@ def venv_activate_render(
...
         sw_cntr_content = host_content
 
-    _render_sw_activate(sw_cntr_content, host_content, workdir)
+    _render_sw_activate(sw_cntr_content, host_content, workdir, quiet)
 
 
-def _render_sw_activate(sw_cntr_content: str, host_content: str, workdir: Path) -> None:
+def _render_sw_activate(
+    sw_cntr_content: str, host_content: str, workdir: Path, quiet: bool = False
+) -> None:
```
jialeicui, line 455: We use `quiet` to tell if the action should break no not when an error occurs:  https://github.com/star-whale/starwhale/blob/ef181f3f83e2a51fe1f8078942c44a1bf5f376b1/client/starwhale/base/tag.py#L60
Will `verbose` be better?

jialeicui:
LGTM

client/starwhale/utils/venv.py
tianweidut: fixed

jialeicui (approved):

PR #895, opened 2022-08-09
feat(client): add --token argument for swcli instance login command

tianweidut:
## Description
image

## Modules
- [x] Client

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/core/instance/cli.py
+++ b/client/starwhale/core/instance/cli.py
@@ -1,3 +1,4 @@
+import sys
 import typing as t
 
 import click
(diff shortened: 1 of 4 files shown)


client/starwhale/core/instance/view.py
```diff
@@ -43,7 +43,7 @@ def login(self, instance: str, username: str, password: str, alias: str) -> None
         r = requests.post(
             url,
             timeout=DEFAULT_HTTP_TIMEOUT,
-            data={"userName": username, "userPwd": password},
+            data=auth_request,
```
jialeicui, line 46: My understanding is that login API is used to generate tokens, there is no need to send the token to login API
cc @dreamlandliu

client/starwhale/core/instance/cli.py
```diff
@@ -28,15 +29,25 @@ def _select(instance: str) -> None:
 
 @instance_cmd.command("login")
 @click.argument("instance", default="")
-@click.option("--username", prompt="username", required=True)
-@click.password_option(confirmation_prompt=False)
+@click.option("--username")
+@click.option("--password")
+@click.option("--token", help="Get token from Starwhale Cloud page")
```
jialeicui, line 34: Would it have been better if the word `Get` was removed?

jialeicui:
LGTM

client/starwhale/core/instance/cli.py
tianweidut: ok

client/starwhale/core/instance/view.py
tianweidut: I hope someone API can check the token validity and get the user info.

tianweidut: Maybe using `/api/v1/user/current` API to check is the better choice.

tianweidut: @jialeicui  updated

jialeicui (approved):

PR #945, opened 2022-08-19
chore(client): add typing hint for python lambda

tianweidut:
## Modules
- [x] Client

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/api/_impl/model.py
+++ b/client/starwhale/api/_impl/model.py
@@ -31,8 +31,10 @@ from . import wrapper
 
 _TASK_ROOT_DIR = "/var/starwhale" if in_production() else "/tmp/starwhale"
 
-_p = lambda p, sub: Path(p) if p else Path(_TASK_ROOT_DIR) / sub
 _ptype = t.Union[str, None, Path]
+_p: t.Callable[[_ptype, str], Path] = (
+    lambda p, sub: Path(p) if p else Path(_TASK_ROOT_DIR) / sub
+)
 
 
 class _LogType:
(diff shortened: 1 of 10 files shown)


client/starwhale/api/_impl/model.py
```diff
@@ -31,16 +31,20 @@
 
 _TASK_ROOT_DIR = "/var/starwhale" if in_production() else "/tmp/starwhale"
 
-_p = lambda p, sub: Path(p) if p else Path(_TASK_ROOT_DIR) / sub
 _ptype = t.Union[str, None, Path]
+_p: t.Callable[[_ptype, str], Path] = (
+    lambda p, sub: Path(p) if p else Path(_TASK_ROOT_DIR) / sub
+)
```
jialeicui, line 37: Do we really need the type hint for the lambda?  or just use `def` like 
```python
def _p(p: _ptype, sub: str) -> Path:
    return Path(p) if p else Path(_TASK_ROOT_DIR) / sub
``` 
The type follows the variable, making it clearer.
At the same time, I think the `_p` should have a more meaningful name.

jialeicui:
LGTM

client/starwhale/api/_impl/model.py
tianweidut: the original purpose: only need a shortcut, not too well-named alias function.
If not typing hint for lambda, we should add `#typing: ignore` for the lambda-used lines.

jialeicui (approved):

PR #986, opened 2022-08-24
fix(controller): make resource pool feature compatible with the old data

jialeicui:
## Description

It may get NPE when the new program runs with the old data.

## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


anda-ren (approved):
lgtm

PR #1010, opened 2022-08-25
fix(controller): prevent losing data after finding the eval_jobs.yaml file

jialeicui:
## Description

The original InputStream has the wrong position after getting the `eval_jobs.yaml` file.
And The InputStream from MultipartFile.getInputStream did not implement the mark func.

This PR is just a walk round for the next integration tests, it will use double memory.

cc @tianweidut 

## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


anda-ren (approved):
lgtm

PR #1011, opened 2022-08-26
fix(dataset): add default s3 region for s3 link auth

tianweidut:
## Modules
- [x] Python-SDK

## Checklist
- [ ] add unit test
- [ ] add necessary doc


anda-ren (approved):
lgtm

PR #1044, opened 2022-08-29
fix(datastore): close inputstream after using s3.get

jialeicui:
## Description

Close `InputStream` after using s3.get


anda-ren (approved):
lgtm

Issue #925, opened 2022-08-18
add `--runtime` for dataset build


tianweidut:
**Current Behavior**
Use the current activated shell environment to build the dataset.

**Proposed Behavior**
- add `--runtime`, `--runtime-restore` argument for dataset building
- record runtime uri in the _manifest.yaml

tianweidut:
https://github.com/star-whale/starwhale/pull/1062

PR #1062, opened 2022-08-29
feat(dataset): support dataset build to use the specific runtime environment with `--runtime` argument

tianweidut:
## Description
- related: https://github.com/star-whale/starwhale/issues/925
image

## Modules
- [x] Client

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/core/dataset/cli.py
+++ b/client/starwhale/core/dataset/cli.py
@@ -29,6 +29,8 @@ def dataset_cmd(ctx: click.Context) -> None:
 )
 @click.option("--append", is_flag=True, help="Only append new data")
 @click.option("--append-from", default=LATEST_TAG, help="Append from dataset version")
+@click.option("--runtime", default="", help="runtime uri")
+@click.option("--runtime-restore", is_flag=True, help="Force to restore runtime")
 @click.pass_obj
 def _build(
     view: DatasetTermView,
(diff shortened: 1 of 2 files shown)


client/starwhale/core/dataset/cli.py
```diff
@@ -29,6 +29,8 @@ def dataset_cmd(ctx: click.Context) -> None:
 )
 @click.option("--append", is_flag=True, help="Only append new data")
 @click.option("--append-from", default=LATEST_TAG, help="Append from dataset version")
+@click.option("--runtime", default="", help="runtime uri")
+@click.option("--runtime-restore", is_flag=True, help="Force to restore runtime")
```
jialeicui, line 33: Do users want to restore the runtime env when they are building a dataset?

jialeicui:
LGTM

jialeicui (approved):

client/starwhale/core/dataset/cli.py
tianweidut: `runtime-restore`  scenario: 
- Users want to build the dataset from one standalone runtime URI.
- But the runtime's workdir is empty, in another word the runtime cannot be activated directly. Then create a virtual environment and install dependencies by restoring command. Activate it when run dataset build.

`runtime-restore` command can be executed repeatedly and fastly with the cache which is a shortcut argument for model/dataset/eval that is using the runtime.  If ignore `runtime-restore` argument, users should run `swcli runtime restore` command in advance.

jialeicui: I get id :ok_hand:

Issue #1063, opened 2022-08-30
tune `--runtime` argument for mode/dataset/eval command


tianweidut:
**Current Behavior**
Users should specify `--runtime-restore` by manual.

**Proposed Behavior**
Restore runtime automatically.

cc @jialeicui

PR #1064, opened 2022-08-30
feat(model): support `--runtime` argument for model build and auto restore runtime

tianweidut:
## Description
- add `--runtime` for the model build, the reason is gen_steps will import user [code](https://github.com/star-whale/starwhale/blob/main/client/starwhale/api/_impl/job.py#L154).
- tune `--runtime-restore` argument, when detecting the venv or conda dir is invalid, the command will restore runtime automatically.
- related: https://github.com/star-whale/starwhale/issues/1063

## Modules
- [x] Client

## Checklist
- [ ] add unit test
- [ ] add necessary doc


jialeicui (approved):
LGTM

PR #1065, opened 2022-08-30
fix(docker): fix wrong python version string parsed by awk

jialeicui:
## Description

Use real yaml processor instead of raw string processing

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [x] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


jialeicui:
#1075

PR #1127, opened 2022-09-06
feat(controller): add file system support for StorageAccessService

xuchuan:
## Modules
- [ ] UI
- [X] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [X] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/resources/application.yaml
+++ b/server/controller/src/main/resources/application.yaml
@@ -34,6 +34,7 @@ sw:
       job-template-path: ${SW_K8S_JOB_TEMPLATE_PATH:}
   storage:
     path-prefix: ${SW_STORAGE_PREFIX:StarWhale}
+    fs-root-dir: ${SW_STORAGE_FS_ROOT_DIR:/usr/local/starwhale}
     s3-config:
       bucket: ${SW_STORAGE_BUCKET:starwhale}
       accessKey: ${SW_STORAGE_ACCESSKEY:starwhale}
--- a/server/storage-access-layer/pom.xml
+++ b/server/storage-access-layer/pom.xml
@@ -7,8 +7,8 @@
   -->
 
 <project xmlns="http://maven.apache.org/POM/4.0.0"
-  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <parent>
         <artifactId>server</artifactId>
         <groupId>ai.starwhale</groupId>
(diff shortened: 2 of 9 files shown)


server/controller/src/main/resources/application.yaml
```diff
@@ -34,6 +34,7 @@ sw:
       job-template-path: ${SW_K8S_JOB_TEMPLATE_PATH:}
   storage:
     path-prefix: ${SW_STORAGE_PREFIX:StarWhale}
+    fs-root-dir: ${SW_STORAGE_FS_ROOT_DIR:.}
```
goldenxinxing, line 37: How about using a fixed directory like “/opt/starwhale/storage”?

server/storage-access-layer/pom.xml
```diff
@@ -36,6 +36,21 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest</artifactId>
+            <version>2.2</version>
```
goldenxinxing, line 51: this version can managed by parent pom at block of "dependencyManagement", and  just use it without write version at childern pom

server/storage-access-layer/pom.xml
```diff
@@ -49,6 +64,20 @@
             </dependency>
         </dependencies>
     </dependencyManagement>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
```
goldenxinxing, line 71: this version can also managed by parent pom at "pluginManagement"

server/controller/src/main/resources/application.yaml
xuchuan: Done

server/storage-access-layer/pom.xml
xuchuan: Done

server/storage-access-layer/pom.xml
xuchuan: Done

server/storage-access-layer/src/main/java/ai/starwhale/mlops/storage/fs/StorageAccessServiceFile.java
```diff
@@ -0,0 +1,136 @@
...
+
+    @Override
+    public void put(String path, InputStream inputStream, long size) throws IOException {
+        var f = new File(this.rootDir, path);
+        //noinspection ResultOfMethodCallIgnored
+        f.getParentFile().mkdirs();
+        var temp = File.createTempFile("sw_tmp", null);
+        try (var out = new FileOutputStream(temp)) {
+            inputStream.transferTo(out);
+            out.close();
```
anda-ren, line 76: `The try-with-resources statement ensures that each resource is closed at the end of the statement. Any object that implements java.lang.AutoCloseable, which includes all objects which implement java.io.Closeable, can be used as a resource`
https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

goldenxinxing: yes，so `out.close` could remove from here

xuchuan: There is a move operation followed by. Although it is not a problem on Linux, I guess it is a good idea to close the file before any further operation.

goldenxinxing (approved):
LGTM

PR #1195, opened 2022-09-14
chore(datastore): use explicit typing in query instead of guessing type from actual column value

xuchuan:
## Modules
- [ ] UI
- [X] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [X] add unit test
- [ ] add necessary doc


anda-ren (approved):
lgtm

Issue #1228, opened 2022-09-16
Evaluation results problem


lijing-susan:
**Describe the bug**
![image](https://user-images.githubusercontent.com/101299635/190572426-bc48f0a7-96e2-4d7d-9a09-a81adfa551fd.png)

**To Reproduce**

**Expected behavior**

**Environment**

PR #1231, opened 2022-09-16
fix(datastore): datastore table list out of bound error when invoke scan api

goldenxinxing:
## Description
close: #1228 
## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


jialeicui:
Does query API need the same logic?

goldenxinxing:
> Does query API need the same logic?

already added this logic in ai.starwhale.mlops.datastore.DataStore#query

jialeicui (approved):
LGTM

PR #1254, opened 2022-09-20
enhance(runtime): runtime quickstart will create workdir automatically

tianweidut:
## Modules
- [x] Client

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/core/runtime/model.py
+++ b/client/starwhale/core/runtime/model.py
@@ -752,12 +752,15 @@ class StandaloneRuntime(Runtime, LocalStorageBundleMixin):
     @classmethod
     def quickstart_from_uri(
         cls,
-        workdir: Path,
+        workdir: t.Union[Path, str],
         name: str,
         uri: URI,
         force: bool = False,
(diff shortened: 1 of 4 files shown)


client/starwhale/core/runtime/model.py
```diff
@@ -752,12 +752,15 @@ def list(
     @classmethod
     def quickstart_from_uri(
         cls,
-        workdir: Path,
+        workdir: t.Union[Path, str],
```
goldenxinxing, line 755: is it always a PATH type? because in 'starwhale.core.runtime.cli._quickstart_from_uri'  has deal with 'p_workdir = Path(workdir).absolute()'

tianweidut: Core model's function only concerns the handlable types which don't know the client's behavior.

goldenxinxing: ok

goldenxinxing (approved):
LGTM

PR #1256, opened 2022-09-20
fix(controller): oss list all keys

jialeicui:
## Description

**AWS S3**
https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html#API_ListObjectsV2_RequestSyntax
>  By default the action returns up to 1,000 key names.
   The response might contain fewer keys but will never contain more

**aliyun oss**
https://github.com/aliyun/aliyun-oss-java-sdk/blob/a927e670955350ff78e1d0b58cb21538923ebf83/src/main/java/com/aliyun/oss/model/ListObjectsRequest.java#L158-L170
```java
/**
 * Sets the max objects to return. By default it's 100, the max is 1000.
 * 
 * @param maxKeys
 *            The max objects to return. The max value is 1000.
 */
public void setMaxKeys(Integer maxKeys) {
    if (maxKeys < 0 || maxKeys > MAX_RETURNED_KEYS_LIMIT) {
        throw new IllegalArgumentException(OSSUtils.OSS_RESOURCE_MANAGER.getString("MaxKeysOutOfRange"));
    }

    this.maxKeys = maxKeys;
}
```
**MinIO**
MinIO SDK does not have this issue.

## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/storage-access-layer/src/main/java/ai/starwhale/mlops/storage/aliyun/StorageAccessServiceAliyun.java
+++ b/server/storage-access-layer/src/main/java/ai/starwhale/mlops/storage/aliyun/StorageAccessServiceAliyun.java
@@ -27,11 +27,15 @@ import com.aliyun.oss.OSSErrorCode;
 import com.aliyun.oss.OSSException;
 import com.aliyun.oss.model.GetObjectRequest;
 import com.aliyun.oss.model.HeadObjectRequest;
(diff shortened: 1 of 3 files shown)


anda-ren (approved):
lgtm

PR #1261, opened 2022-09-20
feat(client): support force remove for model/dataset/runtime/eval

tianweidut:
## Description
use `--force` to remove objects directly which cannot recover.

## Modules
- [x] Client

## Checklist
- [x] add unit test
- [ ] add necessary doc


goldenxinxing (approved):
LGTM

PR #1286, opened 2022-09-22
ci: add auto publish script

anda-ren:
## Description
Usage example:
```
$ cd starwhale/scripts/publish
$ bash pub.sh

Usage: pub.sh [OPTIONS] OBJECT [ARGS]...
Options:
  --help        Show this message and exit.
  --config      config will overwrite your local~/.pypirc and ~/.pip/pip.conf and do a docker login to nexus so that following function will work
OBJECT:
  cli           Publish cli pypi version to nexus & publish runtime base image to nexus
  console       Publish console to k8s
  controller    Update controller to k8s only in dev use case
  all           Publish both cli and console

```
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [x] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


tianweidut (approved):
LGTM

PR #1238, opened 2022-09-19
enhance(console): support getting token for swcli login

jialeicui:
## Description
cc @lijing-susan 

trigger url:  https://my.foo.bar?token=true

https://user-images.githubusercontent.com/3217223/190981295-15a0e261-dd23-4fa5-94e0-32a22c94d438.mp4

## Modules
- [x] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] run code format and lint check
- [ ] add unit test
- [ ] add necessary doc

PR #1309, opened 2022-09-28
fix(console): change token key in url param

jialeicui:
## Description

`token` may conflict with creating account logic
trigger URL will change to: https://my.foo.bar/?cli-token=true
Relates to #1238

cc @lijing-susan 

## Modules
- [x] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


waynelwz (approved):
LGTM

PR #1320, opened 2022-09-29
fix(e2e):optimise e2e

goldenxinxing:
## Description

1. Replace mnist with simple example in e2e test
2. replace 'run_demo.sh' with python script
3. Print the task log regardless of success or failure
4. fix some tiny bug
- get_data_loader(): end+1 -> end
- PipeHandler repeat record ppl/cmp's status
- add flush option for ppl（an optional that default to false for use in some single-thread unit test without atexit）

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [x] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


anda-ren (approved):
lgtm

PR #1351, opened 2022-10-13
fix(client): unit test works with macos

jialeicui:
## Description
Unit test works with macOS
* symlink can not work with virtualenv (not only in macOS)
* temp dir under macOS is like `/var/folders/xx/xxxxxxxxxxxxxxxx/X/`, not `/tmp/`
* macOS use a symlink temp dir for users

mnist demo can not run under macOS for now, will fix it in further PR.

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [x] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/.github/workflows/client.yaml
+++ b/.github/workflows/client.yaml
@@ -90,8 +90,8 @@ jobs:
           - "3.9"
           - "3.10.6" # https://github.com/python/mypy/issues/13627
         os:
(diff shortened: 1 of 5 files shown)


tianweidut (approved):
LGTM

PR #1395, opened 2022-10-19
fix(client): get instance wrong when multi instances has the same uri

goldenxinxing:
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/utils/config.py
+++ b/client/starwhale/utils/config.py
@@ -158,9 +158,18 @@ class SWCliConfigMixed:
             return self.current_instance
 
         if instance not in self._config["instances"]:
+            _count = 0
+            _alias = None
             for k, v in self._config["instances"].items():
                 if v["uri"] == instance:
-                    return str(k)
+                    _alias = str(k)
+                    _count += 1
+            if _count > 1:
+                raise RuntimeError(
+                    f"instance uri:{instance} has multi items!! please use alias."
+                )
+            if _alias:
+                return _alias
 
         return instance
 
--- a/client/tests/base/test_uri.py
+++ b/client/tests/base/test_uri.py
@@ -157,3 +157,6 @@ class URITestCase(TestCase):
 
         uri = URI("", expected_type=URIType.PROJECT)
         assert uri.project == "self"
+
+        with self.assertRaises(RuntimeError):
+            URI("http://2.2.2.2:8182", expected_type=URIType.INSTANCE)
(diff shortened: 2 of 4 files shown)


client/starwhale/base/uri.py
```diff
@@ -216,7 +216,11 @@ def real_request_uri(self) -> t.Union[str, Path]:
 
     @property
     def sw_instance_config(self) -> t.Dict[str, t.Any]:
-        return self._sw_config.get_sw_instance_config(self.instance)
+        return (
+            self._sw_config.get_sw_instance_config(self.instance)
+            if not self.instance_alias
```
jialeicui, line 221: It seems that the `instance_alias` will always have a value.
https://github.com/star-whale/starwhale/blob/5abf68e650c66ebceed81c18f52d48ee8c6f39ed/client/starwhale/base/uri.py#L106-L109

client/starwhale/base/uri.py
```diff
@@ -216,7 +216,11 @@ def real_request_uri(self) -> t.Union[str, Path]:
 
     @property
     def sw_instance_config(self) -> t.Dict[str, t.Any]:
-        return self._sw_config.get_sw_instance_config(self.instance)
+        return (
+            self._sw_config.get_sw_instance_config(self.instance)
```
jialeicui, line 220: How do we define the situation when a user uses the same URL with a different token?
cc @tianweidut

tianweidut: a URL + a token = a instance

goldenxinxing: a URL + a token == an instance alias?

goldenxinxing: if user write uri with the foramt of '${instance URL}/project/...', we can only filter the instance from config by url, but if the format is '${instance-alias}/project/...', We just to use alias to find exactly what the user is referring to。

client/starwhale/base/uri.py
goldenxinxing: done

client/starwhale/utils/config.py
```diff
@@ -147,8 +147,13 @@ def current_instance(self) -> str:
         return str(self._config["current_instance"])
 
     def get_sw_instance_config(self, instance: str) -> t.Dict[str, t.Any]:
-        instance = self._get_instance_alias(instance)
-        return self._config["instances"].get(instance) or {}
+        instance_alias = self._get_instance_alias(instance)
+        return self._config["instances"].get(instance_alias) or {}
```
jialeicui, line 151: `return get_sw_instance_config_by_alias(instance_alias)` ?

client/starwhale/utils/config.py
```diff
@@ -147,8 +147,13 @@ def current_instance(self) -> str:
         return str(self._config["current_instance"])
 
     def get_sw_instance_config(self, instance: str) -> t.Dict[str, t.Any]:
-        instance = self._get_instance_alias(instance)
-        return self._config["instances"].get(instance) or {}
+        instance_alias = self._get_instance_alias(instance)
```
jialeicui, line 150: `_get_instance_alias` should throw an exception when getting duplicate urls?

goldenxinxing: done

jialeicui (approved):
LGTM

PR #1468, opened 2022-11-08
feat(controller): add temporary token for task

anda-ren:
## Description
- token is bind with task id when doing a remote task.
- when task is not running, this token is not valid.
- the expire time of token is the same with system default

verify environment http://task-token-20221108.pre.intra.starwhale.ai/
## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc


jialeicui (approved):
LGTM

dreamlandliu (approved):
lgtm

PR #1486, opened 2022-11-11
enhance(datastore): add map type support

xuchuan:
## Modules
- [ ] UI
- [X] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [X] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/java/ai/starwhale/mlops/datastore/ColumnSchemaDesc.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/datastore/ColumnSchemaDesc.java
@@ -54,6 +54,8 @@ public class ColumnSchemaDesc {
      * <li>STRING</li>
      * <li>BYTES</li>
      * <li>LIST</li>
+     * <li>TUPLE</li>
+     * <li>MAP</li>
      * <li>OBJECT</li>
      * </ul>
      */
@@ -66,10 +68,22 @@ public class ColumnSchemaDesc {
     private String pythonType;
 
     /**
-     * This field represents the type of elements in array. It is only used when type is ARRAY. The name field of
+     * This field represents the type of elements in the list. It is only used when type is LIST. The name field of
      * elementType is never used.
      */
     private ColumnSchemaDesc elementType;
+
+    /**
+     * This field represents the type of keys in the map. It is only used when type is MAP. The name field of
+     * keyType is never used.
+     */
+    private ColumnSchemaDesc keyType;
+
+    /**
+     * This field represents the type of values in map. It is only used when type is MAP. The name field of
+     * valueType is never used.
+     */
+    private ColumnSchemaDesc valueType;
 
     /**
      * This field describes the attributes of object type.
--- a/server/controller/src/main/java/ai/starwhale/mlops/datastore/ColumnType.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/datastore/ColumnType.java
@@ -22,6 +22,7 @@ import java.nio.ByteBuffer;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Stream;
 import lombok.Getter;
 import lombok.NonNull;
 import org.apache.parquet.io.api.Converter;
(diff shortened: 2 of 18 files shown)


server/controller/src/main/java/ai/starwhale/mlops/datastore/ColumnType.java
```diff
@@ -89,102 +101,168 @@ public static int compare(ColumnType type1, Object value1, ColumnType type2, Obj
...
-                            long v1 = (Long) value1;
-                            if (v1 > (1L << 53)) {
-                                return BigDecimal.valueOf(v1)
-                                        .compareTo(BigDecimal.valueOf(((Number) value2).doubleValue()));
-                            }
+                if (category1.equals(ColumnTypeScalar.INT32.getCategory())
+                        && category2.equals(ColumnTypeScalar.FLOAT32.getCategory())) {
+                    if (type1 == ColumnTypeScalar.INT64) {
+                        long v1 = (Long) value1;
+                        if (v1 > (1L << 53)) {
```
goldenxinxing, line 150: so, considering the precision of js, can't we store more than 53 digits? shouldn't such precision issues be considered in server?

server/controller/src/main/java/ai/starwhale/mlops/datastore/ColumnSchemaDesc.java
```diff
@@ -66,11 +66,23 @@ public class ColumnSchemaDesc {
...
+     * This field represents the type of keys in the map. It is only used when type is MAP. The name field of
+     * keyType is never used.
+     */
+    private ColumnSchemaDesc keyType;
+
+    /**
+     * This field represents the type of values in map. It is only used when type is MAP. The name field of
+     * valueType is never used.
+     */
+    private ColumnSchemaDesc valueType;
```
jialeicui, line 84: Add `MAP` in the comment of `private String type;`

server/controller/src/main/java/ai/starwhale/mlops/datastore/ColumnType.java
```diff
@@ -89,102 +101,168 @@ public static int compare(ColumnType type1, Object value1, ColumnType type2, Obj
...
+        int result = 0;
+        for (var iter = Stream.concat(map1.keySet().stream(), map2.keySet().stream()).iterator();
+                iter.hasNext(); ) {
+            var key = iter.next();
+            if (minDiffKey == null || ColumnType.compare(type1.getKeyType(), key, type1.getKeyType(), minDiffKey) < 0) {
+                var v1 = map1.get(key);
+                var v2 = map2.get(key);
+                int t;
+                if (v1 == null) {
+                    t = 1;
```
jialeicui, line 209: -1?

server/controller/src/main/java/ai/starwhale/mlops/datastore/ColumnType.java
```diff
@@ -89,102 +101,168 @@ public static int compare(ColumnType type1, Object value1, ColumnType type2, Obj
...
+        var map2 = (Map<?, ?>) value2;
+        if (map1.isEmpty() && map2.isEmpty()) {
+            return 0;
+        }
+        if (map1.isEmpty()) {
+            return -1;
+        }
+        if (map2.isEmpty()) {
+            return 1;
+        }
```
jialeicui, line 198: This block is just a fast path? can we remove it?

server/controller/src/main/java/ai/starwhale/mlops/datastore/ColumnType.java
xuchuan: Here is the logic of comparing an int64 with a float. Integers greater than 1^53 can not be converted to Double without precision loss. That's why here is a if statement. BigDecimal is used if the value is greater than 1^53, or Double otherwise.

server/controller/src/main/java/ai/starwhale/mlops/datastore/ColumnType.java
```diff
@@ -89,102 +101,168 @@ public static int compare(ColumnType type1, Object value1, ColumnType type2, Obj
...
+        Object minDiffKey = null;
+        int result = 0;
+        for (var iter = Stream.concat(map1.keySet().stream(), map2.keySet().stream()).iterator();
```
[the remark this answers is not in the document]
xuchuan: not necessary. distinct is redundant here. The same key will be quickly skipped because it will not satisfy the condition "< minDiffKey"

server/controller/src/main/java/ai/starwhale/mlops/datastore/ColumnType.java
xuchuan: Changed to another algorithm implementation.

server/controller/src/main/java/ai/starwhale/mlops/datastore/ColumnType.java
xuchuan: Changed to another algorithm implementation.

server/controller/src/main/java/ai/starwhale/mlops/datastore/ColumnSchemaDesc.java
xuchuan: Done

jialeicui (approved):
LGTM

PR #1532, opened 2022-11-18
enhance(sdk): tune get http file performance with buffer

tianweidut:
## Description
- fix unclosed issue for _make_file
- support HttpBufferedFileLike same as `S3BufferedFileLike`
- tune HTTP get file with timeout, stream and buffered.

## Modules
- [x] Client
- [x] Python-SDK

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/core/dataset/store.py
+++ b/client/starwhale/core/dataset/store.py
@@ -7,12 +7,12 @@ import json
 import shutil
 import typing as t
 from abc import ABCMeta, abstractmethod
+from types import TracebackType
 from pathlib import Path
 from urllib.parse import urlparse
 
 import boto3
 import requests
-from loguru import logger
 from botocore.client import Config as S3Config
 from typing_extensions import Protocol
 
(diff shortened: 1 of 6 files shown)


anda-ren (approved):
lgtm

client/starwhale/core/dataset/store.py
```diff
@@ -438,19 +453,151 @@ def sign_uri(self, uri: str) -> str:
...
+        if size < 0 or size > len(self):
+            size = len(self)
+
+        ret = self._buffer[self._pos : self._pos + size]
+        self._pos += size
+        return ret.tobytes()
+
+    def write(self, new_bytes: bytes) -> int:
+        _release_buffer = self._buffer
+        self._buffer = memoryview(self._buffer[self._pos :].tobytes() + new_bytes)
```
jialeicui, line 576: `self._buffer[self._pos :].tobytes() + new_bytes` will allocate new memory and create another copy of the data, is this expected?

client/starwhale/core/dataset/store.py
```diff
@@ -438,19 +453,151 @@ def sign_uri(self, uri: str) -> str:
...
+            timeout=timeout,
+        )
+
+        if not self.resp.ok:
+            self.resp.raise_for_status()
+
+        self._read_iter = self.resp.iter_content(buffer_size)
+
+    def _read_exhausted_raw(self) -> bytes:
+        return self.resp.raw.read()  # type: ignore
```
jialeicui, line 539: Will it be better that `HttpBufferedFileLike` overrides the `read` function and just return self.resp.raw.read(size)?

tianweidut: one read case:

```
read(32)  <-- first read 32 bytes
read(8)    <-- then read 8 bytes
read(-1)   <--  last read all remaining bytes
```
from the bellowing example, some bytes already maybe read into the buffer before , so we cannot read `self.resp.raw.read` directly.

Another we use the buffer for http stream read.

tianweidut: updated

client/starwhale/core/dataset/store.py
tianweidut: updated,  use `bytearray` directly. However, sometimes write function needs allocate new memory.

jialeicui (approved):
LGTM

PR #1548, opened 2022-11-22
chore(datastore): return Iterator in MemoryTable.query

xuchuan:
## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/java/ai/starwhale/mlops/datastore/impl/MemoryTableImpl.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/datastore/impl/MemoryTableImpl.java
@@ -45,6 +45,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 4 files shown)


server/controller/src/main/java/ai/starwhale/mlops/datastore/impl/MemoryTableImpl.java
```diff
@@ -361,20 +346,7 @@ public List<RecordResult> query(
                 return 0;
             });
         }
-
-        return results.subList(start, end).stream().map(record -> {
```
jialeicui, line 365: What are the benefits of moving (start, end) logic outside?

xuchuan: In order to support versions, DataStore.query should be able to merge multiple tables just like scan. start & limit only apply to the merged result so that corresponding parameters in MemoryTable.query become meaningless.

jialeicui: OK, I get it.

jialeicui (approved):
LGTM

PR #1549, opened 2022-11-23
enhance(datastore): support server side flush

xuchuan:
## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/java/ai/starwhale/mlops/datastore/WalManager.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/datastore/WalManager.java
@@ -47,6 +47,8 @@ public class WalManager extends Thread {
 
     private static final int WAIT_INTERVAL_MILLIS = 1000;
 
+    private static final long FLUSH_ID = -12345;
+
     private final StorageAccessService storageAccessService;
 
     private final int walFileSize;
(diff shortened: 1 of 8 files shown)


server/controller/src/main/java/ai/starwhale/mlops/datastore/WalManager.java
```diff
@@ -385,6 +420,13 @@ private PopulationStatus populateOutput() {
                 }
                 entry = this.entriesToWrite.getFirst();
             }
+            if (entry.getId() == FLUSH_ID) {
+                synchronized (this.entriesToWrite) {
+                    this.entriesToWrite.removeFirst();
```
jialeicui, line 425: It may delete the unexpected entry?

xuchuan: FLUSH_ID = -12345, which will never appear in normal entries.

jialeicui: I didn't make myself clear, I mean that `getFirst` and `removeFirst` are not in the same synchronized block.

xuchuan: entriesToWrite is a queue. Entries are removed from the front and appended to the back. Only populateOutput may remove entries, so it is not a problem.

jialeicui (approved):
LGTM

PR #1565, opened 2022-11-30
feat(sdk): support online serving

jialeicui:
## Description

SDK support online serving

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [x] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [x] add necessary doc


client/starwhale/api/_impl/service.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class Response(ABC):
+    # TODO: support mime_type and body
+    @abstractmethod
+    def dump(self, resp: t.Any) -> bytes:
+        raise NotImplementedError()
+
+
+class GrayscaleImageRequest(Request):
```
tianweidut, line 23: can we move the load function into [dataset/type.py](https://github.com/star-whale/starwhale/blob/main/client/starwhale/core/dataset/type.py) `GrayscaleImage` class?  From the user's view, the `GrayscaleImage`  can build a dataset and receive HTTP requests.

tianweidut:
```
from typing_extensions import Protocol
class Request(Protocol):
    def load(self, req: t.Any) -> t.Any:
         ...
class Response(Protocol):
    def dump(self, req: t.Any) -> t.Any:
         ...
```
can we use `Protocol` to define `Request` and `Response`?

client/starwhale/core/model/view.py
```diff
@@ -145,6 +148,19 @@ def tag(self, tags: t.List[str], remove: bool = False, quiet: bool = False) -> N
             console.print(f":surfer: add tags [red]{tags}[/] @ {self.uri}...")
             self.model.add_tags(tags, quiet)
 
+    @classmethod
+    def serve(
```
tianweidut, line 152: add `@BaseTermView._only_standalone` for serve function?

tianweidut (commented):
Others LGTM

client/starwhale/api/_impl/service.py
jialeicui: Done

client/starwhale/core/model/view.py
jialeicui: Done

tianweidut (approved):
LGTM

PR #1569, opened 2022-12-01
feat(sdk): allow json like dict passed as dataset annotation

anda-ren:
## Description
This PR allows user to pass any `dict` loaded from JSON to dataset index annotations excpet who's values in arry don't have the same type. `SwObject` is an added valid value type besides JSON standard.
#### Valid annotation
Before this PR the annotation below is an invalid input, because `c.ca` and `c.cb` have different types. This PR makes it valid.
```python
{
	"a": 1,
	"b": [1, 2, 3],
	"c": {
                "ca": ["1","2","3"],
		"cb": 1
	},
	"d": Link("http://ad.c/d")
}
```
#### Invalid annotation
This annotation below is still invalid, because `b`'s values have different types. If it needs to be valid,  `SwList`  should be modified.
```python
{
	"a": 1,
	"b": [1, "2", {"a":"b"}],
	"c": {
		"ca": "1"
	},
	"d": Link("http://ad.c/d")
}
```

## Change list

1. Add `JsonDict` class inherited from `SwObject`
2. Convert annotations to `JsonDict` when saving them to datastore
3. Convert `JsonDict` to  `dict` when loading them from datastore
4. Adapt UI to this change
5. Change coco dataset example to adopt this feature

## Validation
http://json-dict-20221201.pre.intra.starwhale.ai
image
image
image

## Modules
- [x] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc


waynelwz (dismissed):
LGTM

client/starwhale/core/dataset/type.py
```diff
@@ -702,3 +702,53 @@ def create_by_yaml(cls, fpath: t.Union[str, Path]) -> DatasetConfig:
...
+
+class JsonDict(SwObject):
+    """
+    JsonDict takes json like dict(https://www.json.org/json-en.html) as init parameter
+    Besides the standard value types, SwObject is an extra value type that is allowed to be passed in
+    """
+
+    def __init__(self, d: dict = {}) -> None:
+        for _k, _v in d.items():
+            if type(_k) != str:
```
tianweidut, line 715: why the key type must be str type?

```python
In [1]: a = {1: "test"}

In [2]: type(a)
Out[2]: dict
```

client/starwhale/core/dataset/type.py
```diff
@@ -702,3 +702,53 @@ def create_by_yaml(cls, fpath: t.Union[str, Path]) -> DatasetConfig:
...
+                raise ValueError("json like dict shouldn't have none-str keys")
```
tianweidut, line 716: add `_k` value into the exception message? it can help users to find out the exception reason.

client/starwhale/core/dataset/type.py
```diff
@@ -702,3 +702,53 @@ def create_by_yaml(cls, fpath: t.Union[str, Path]) -> DatasetConfig:
...
+            return cls(d)
+        for _t, _ in _TYPE_DICT.items():
+            if isinstance(d, _t):
+                return d
+        if isinstance(d, SwObject):
+            return d
+        if type(d) == list:
+            return [cls.from_data(_d) for _d in d]
+
+        raise ValueError(
```
tianweidut, line 735: `tuple`, `collections.namedtuple` and `set` are also common types.

client/starwhale/core/dataset/type.py
```diff
@@ -702,3 +702,53 @@ def create_by_yaml(cls, fpath: t.Union[str, Path]) -> DatasetConfig:
...
+            "json like dict shouldn't have values who's type is not in [SwObject, list, dict, str, int, bool, None]"
```
tianweidut, line 736: add `d` value into the exception message?

client/starwhale/core/dataset/type.py
```diff
@@ -702,3 +702,53 @@ def create_by_yaml(cls, fpath: t.Union[str, Path]) -> DatasetConfig:
...
+        return ret
+
+    @classmethod
+    def to_data(cls, d: t.Any) -> t.Any:
+        """
+        unwrapp JsonDict to dict
+        """
+        if isinstance(d, JsonDict):
+            return d.asdict()
+        if isinstance(d, list):
```
tianweidut, line 752: add tuple support?

client/starwhale/core/dataset/type.py
```diff
@@ -702,3 +702,53 @@ def create_by_yaml(cls, fpath: t.Union[str, Path]) -> DatasetConfig:
...
+        for k, v in self.__dict__.items():
+            ret[k] = JsonDict.to_data(v)
+        return ret
+
+    @classmethod
+    def to_data(cls, d: t.Any) -> t.Any:
+        """
+        unwrapp JsonDict to dict
+        """
+        if isinstance(d, JsonDict):
```
tianweidut, line 750: can we unpack these schema?

```
{
"a": JsonDict()
}
```

client/starwhale/core/dataset/type.py
anda-ren: Technically, this is not a full python dict support. As the PR's title indicates, this is a subset of python dict which could be called `JSON like` dict. Keys of JSON should be str type. 
From user's aspect, I can think of little cases where non-str keys are used.
image

client/starwhale/core/dataset/type.py
anda-ren: As `collections.namedtuple` `set` are not supported by datastore, they are hard to be processed. I will add `tuple` in the implemention.

client/starwhale/core/dataset/type.py
anda-ren: done

client/starwhale/core/dataset/type.py
anda-ren: done

client/starwhale/core/dataset/type.py
anda-ren: done

client/starwhale/core/dataset/type.py
anda-ren: `tuple` added

client/starwhale/core/dataset/type.py
anda-ren: Yes we can. UT for this case is added.
https://github.com/star-whale/starwhale/blob/c212eabe571c90649ca33ba88bc571e9146bba19/client/tests/core/test_dataset.py#L466

client/starwhale/core/dataset/type.py
tianweidut: ok

client/starwhale/core/dataset/type.py
tianweidut: ok

client/starwhale/__init__.py
```diff
@@ -10,6 +10,7 @@
     Image,
     Video,
     Binary,
+    JsonDict,
```
tianweidut, line 13: Is `JsonDict` a common type for users or is it an internal type for starwhale developers?

The current annotations can support users defining a python dict type,  so the `JsonDict` is transparent for the users.

tianweidut:
Others LGTM

client/starwhale/__init__.py
anda-ren: Yes, the `JsonDict` is transparent for the users currentlly . The exportation is removed in api module.

tianweidut (approved):

PR #1572, opened 2022-12-02
fix(client): prevent module cache and confliction

jialeicui:
### Prevent module cache and confliction

Known premises:
1. When we use `importlib.import_module` to import modules, it uses cached modules in `sys.modules`
2. import_module `a.b` will make `a` and `a.b` in sys.modules
3. Python has a builtin module named `code`

The following will happen when users build custom models and datasets in the same src py:
1. build model
    - copy src to temp dir
    - import the user's custom model via `importlib.import_module`
    - mv temp dir to snapshot dir
    the above actions make the model (`parent` and `parent.model`) in the cache but the src is missing
2.  build dataset
    - try importing `parent.dataset`  (this will cause a not found exception)

So this PR
- import module from work dir instead of temp dir
- rename `code` to `src` to prevent confliction with the builtin `code` module 

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


tianweidut (approved):
LGTM

tianweidut:
@jialeicui  update pr title?

PR #1638, opened 2022-12-20
example(dataset): add celeba-align example & tune ds.append performance

anda-ren:
## Description
- tune ds.append performance(202,599 rows in 1.5Hour->1.5Min)
- add celeba-align example
Screenshot from 2022-12-20 16-48-15

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


client/starwhale/api/_impl/dataset/model.py
```diff
@@ -159,6 +159,9 @@ def __init__(
         self._writer_lock = threading.Lock()
         self._row_writer: t.Optional[RowWriter] = None
         self.__keys_cache: t.Set[t.Union[int, str]] = set()
+        if self._create_by_append:
+            loader = get_data_loader(_origin_uri)
+            self.__keys_cache = set([row.id for row in loader.tabular_dataset.scan()])
```
tianweidut, line 164: it will be too slow to load all data?

client/starwhale/api/_impl/dataset/model.py
```diff
@@ -410,7 +413,6 @@ def info(self) -> TabularDatasetInfo:
     def flush(self) -> None:
         loader = self._get_data_loader(disable_consumption=True)
         loader.tabular_dataset.flush()
-        self.__keys_cache = set()
```
tianweidut, line 413: why remove?

client/starwhale/api/_impl/dataset/builder.py
```diff
@@ -648,11 +648,12 @@ def update(self, row_item: DataRow) -> None:
         self._raise_run_exception()
         self._queue.put(row_item)
 
-        with self._lock:
-            if self._builder is None:
-                _cls = create_generic_cls(self.__iter__)
-                self._builder = _cls(**self._kw)  # type: ignore
-                self.start()
+        if self._builder is None:
+            with self._lock:
```
tianweidut, line 652: the `lock` context change is useless. Python use GIL, and the lock here will add almost no overhead.

client/starwhale/api/_impl/dataset/model.py
tianweidut: Another point, it will make the init process slower. A lazy load is the better choice.

client/starwhale/api/_impl/wrapper.py
```diff
@@ -37,10 +37,11 @@ def close(self) -> None:
                 raise Exception(*exceptions)
 
     def _fetch_writer(self, table_name: str) -> data_store.TableWriter:
+        writer = self._writers.get(table_name)
```
tianweidut, line 40: The tune code almost gains no performance benefit for python because of python GIL. And it brings the redundant code.

PR #1682, opened 2022-12-29
feat: replace link auth file to link auth config

anda-ren:
## Description
users do not need to provide auth info in link, but instead he should provide the auth info in config
[asciicast](https://asciinema.org/a/ZxMlv3RVQnp1jMR2jdOGUmkPP)

## Gains
- Link could be used out of dataset in the future
- User experience is consistent with controller (config link auth in setting)

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc


client/starwhale/api/_impl/dataset/loader.py
```diff
@@ -120,17 +119,12 @@ def __init__(
...
-                DatasetStorage(self.dataset_uri).snapshot_workdir / AUTH_ENV_FNAME
-            )
-            load_dotenv(auth_env_fpath)
-
     def _get_store(self, row: TabularDatasetRow) -> ObjectStore:
         with self._store_lock:
-            _k = f"{self.dataset_uri}.{row.data_link.scheme}.{row.auth_name}"
+            _up = urlparse(row.data_link.uri)
+            parts = _up.path.lstrip("/").split("/", 1)
+            c_k = row.data_link.uri.replace(parts[-1], "")
```
tianweidut, line 126: use another meaningful variable name(`c_k`)? 
In addition, there both are tmp/private variable(`_up`, `_k`) and normal variable(`parts`, `c_k`), use a uniform pattern?

tianweidut: If `parts[-1]` content is also a part as row.data_link.uri netloc, the replace operation will cause the unexpected result.
for example:
the URI is `http://www.test.com/a/test`, and the c_k is `http://www..com/a/`.

In addition, the query part should also be truncated.

tianweidut:
```
_up = urlparse(row.data_link.uri)
parts = _up.path.lstrip("/").split("/", 1)
c_k = row.data_link.uri.replace(parts[-1], "")
_k = f"{self.dataset_uri}.{c_k}"
```
-->
```
_up = urlparse(row.data_link.uri)
_parts = _up.path.lstrip("/").split("/", 1)
_k = f"{self.dataset_uri}.{_up.netloc}.{parts[0]}
```
?

client/starwhale/core/dataset/store.py
```diff
@@ -187,21 +187,29 @@ def close(self) -> None:
 
 
 class S3Connection:
+    connections_config: t.List[S3Connection] = []
+    init_config_lock = threading.Lock()
+    supported_schemes = set(["s3", "minio", "aliyun", "oss"])
```
tianweidut, line 192: `supported_schemes = set(["s3", "minio", "aliyun", "oss"])`
-->
`supported_schemes = {"s3", "minio", "aliyun", "oss"}`
?

client/starwhale/core/dataset/store.py
```diff
@@ -218,55 +226,88 @@ def __init__(
...
         uri:
             - s3://username:password@127.0.0.1:8000/bucket/key
             - s3://127.0.0.1:8000/bucket/key
         """
-        from .type import S3LinkAuth
-
-        uri = uri.strip()
-        if not uri or not uri.startswith(("s3://", "minio://")):
+        r = urlparse(uri.strip())
+        if not r.scheme or r.scheme not in cls.supported_schemes:
```
tianweidut, line 246: `if not r.scheme or r.scheme not in cls.supported_schemes:`
-->
`if r.scheme not in cls.supported_schemes:`
?

if the URI does not contain schema, the urlparse function will return empty str for the schema attribute.

```
In [20]: _up = urlparse("11/a ")

In [21]: _up
Out[21]: ParseResult(scheme='', netloc='', path='11/a ', params='', query='', fragment='')
```

client/starwhale/core/dataset/store.py
```diff
@@ -218,55 +226,88 @@ def __init__(
...
+                return
+            sw_config = SWCliConfigMixed()
+            link_auths = sw_config.link_auths
+            if not link_auths:
+                return
+            for la in link_auths:
+                if not la.get("type") or type(la.get("type")) != str:
+                    continue
+                if not la.get("type") in cls.supported_schemes:
+                    continue
```
tianweidut, line 284: `if not la.get("type") in cls.supported_schemes:`
-->
`if la.get("type") not in cls.supported_schemes:`
?

client/starwhale/utils/config.py
```diff
@@ -248,3 +257,42 @@ def update_instance(
         )
 
         update_swcli_config(**self._config)
+
+
+@click.group(
```
tianweidut, line 262: move click cli definition into cli.py ?
https://github.com/star-whale/starwhale/tree/main/client/starwhale/cli dir maybe a better choice.

client/starwhale/utils/config.py
```diff
@@ -248,3 +257,42 @@ def update_instance(
...
+
+
+@config_cmd.command("edit", aliases=["e"], help="edit the configuration of swlci")
+def __edit() -> None:
+    _edit()
+
+
+def _edit() -> None:
+    _editor = ""
+    if platform == "linux" or platform == "linux2":
```
tianweidut, line 278: use `sys.platform` is better than `platform` global variable?

client/starwhale/utils/config.py
```diff
@@ -248,3 +257,42 @@ def update_instance(
...
+        _editor = os.environ.get(SWEnv.sw_editor, "")
+        if not _editor:
+            _editor = os.environ.get("EDITOR", "vi")
+    elif platform == "darwin" or platform == "win32":
+        _editor = os.environ.get(SWEnv.sw_editor, "")
+        if not _editor:
+            raise NoSupportError(f"No {SWEnv.sw_editor} is configured")
+    else:
+        raise NoSupportError(f"Current platform is not supported by: {platform}")
```
tianweidut, line 287:
```
    if platform == "linux" or platform == "linux2":
        _editor = os.environ.get(SWEnv.sw_editor, "")
        if not _editor:
            _editor = os.environ.get("EDITOR", "vi")
    elif platform == "darwin" or platform == "win32":
        _editor = os.environ.get(SWEnv.sw_editor, "")
        if not _editor:
            raise NoSupportError(f"No {SWEnv.sw_editor} is configured")
    else:
        raise NoSupportError(f"Current platform is not supported by: {platform}")
```
-->
```
from shutil import which
_editor = os.environ.get(SWEnv.sw_editor) or os.environ.get("EDITOR") or "vi"
if which(_editor) is None:
   raise noSupportError(f"no found {_editor} bin in {platform}")
```
?

client/starwhale/utils/config.py
```diff
@@ -248,3 +257,42 @@ def update_instance(
...
+    path_config = get_swcli_config_path()
+    fd, fname = tempfile.mkstemp()
```
tianweidut, line 289: we should mkstemp at the path_config dir, it will avoid cross-device link issue.
`os.rename` only works when the source and target names are on the same file system.

client/starwhale/utils/config.py
```diff
@@ -248,3 +257,42 @@ def update_instance(
...
+    with open(path_config) as config_f, os.fdopen(fd, "w") as tmpf:
+        for line in config_f:
+            tmpf.write(line)
+        tmpf.close()
+        cmd = _editor + " " + fname
+        _c = subprocess.call(cmd, shell=True)
+        if _c:
+            raise ExistedError("editing file with unexpected failure")
+        os.rename(fname, path_config)
```
jialeicui, line 298: - It would be better if we support `VISUAL` env  https://unix.stackexchange.com/questions/4859/visual-vs-editor-what-s-the-difference
- `swcli config edit` is not a high-frequency operation, `SWEnv.sw_editor` may not be used by most of the users.
- `VISUAL` and `EDITOR` should be supported in `darwin`
- Using a temp file may not be necessary (users my lose "swap detection" features when using vim)

Refer to the implementation of `pip config edit`
https://github.com/pypa/pip/blob/c6783501f16b7ff8c54adf2bcbc1f5f1d11ac699/src/pip/_internal/commands/configuration.py#L225

client/starwhale/utils/config.py
anda-ren: sometime the editor command may be a compose of executable and some options, like:
`code -w` in windows or `/Applications/Sublime.app/Contents/SharedSupport/bin/subl -w` in macos. 
In such cases, `which` may not working as expected. However, the code you provide is really concise.

client/starwhale/utils/config.py
anda-ren: > * `swcli config edit` is not a high-frequency operation, `SWEnv.sw_editor` may not be used by most of the users.
> Refer to the implementation of pip config edit
https://github.com/pypa/pip/blob/c6783501f16b7ff8c54adf2bcbc1f5f1d11ac699/src/pip/_internal/commands/configuration.py#L225

`pip config edit` offers an option for users to customize the editor, which functions the same like `SWEnv.sw_editor`. Users may not change the editor very frequently. So, is environment variable better than command line option?

client/starwhale/api/_impl/dataset/loader.py
anda-ren: > If `parts[-1]` content is also a part as row.data_link.uri netloc, the replace operation will cause the unexpected result. for example: the URI is `http://www.test.com/a/test`, and the c_k is `http://www..com/a/`.
> 
> In addition, the query part should also be truncated.

Eventhough the cache key is wired, it works. Yes, the query part need to be truncated

anda-ren: > ```
> _up = urlparse(row.data_link.uri)
> parts = _up.path.lstrip("/").split("/", 1)
> c_k = row.data_link.uri.replace(parts[-1], "")
> _k = f"{self.dataset_uri}.{c_k}"
> ```
> 
> -->
> 
> ```
> _up = urlparse(row.data_link.uri)
> _parts = _up.path.lstrip("/").split("/", 1)
> _k = f"{self.dataset_uri}.{_up.netloc}.{parts[0]}
> ```
> 
> ?

A problem with `parts[0]` is that URI which has only one path part will have a unique cache key itself. The cache mechnism fails when uri looks like `data_ubyte_0.swds_bin`. If `parts[-1]` is replace with empty string, the cache keys reduce to `{self.dataset_uri}.` and the cache works.

anda-ren: > use another meaningful variable name(`c_k`)? In addition, there both are tmp/private variable(`_up`, `_k`) and normal variable(`parts`, `c_k`), use a uniform pattern?

done

client/starwhale/core/dataset/store.py
anda-ren: > `supported_schemes = set(["s3", "minio", "aliyun", "oss"])` --> `supported_schemes = {"s3", "minio", "aliyun", "oss"}` ?

done

client/starwhale/core/dataset/store.py
anda-ren: > `if not r.scheme or r.scheme not in cls.supported_schemes:` --> `if r.scheme not in cls.supported_schemes:` ?
> 
> if the URI does not contain schema, the urlparse function will return empty str for the schema attribute.
> 
> ```
> In [20]: _up = urlparse("11/a ")
> 
> In [21]: _up
> Out[21]: ParseResult(scheme='', netloc='', path='11/a ', params='', query='', fragment='')
> ```

done

client/starwhale/core/dataset/store.py
anda-ren: > `if not la.get("type") in cls.supported_schemes:` --> `if la.get("type") not in cls.supported_schemes:` ?

done

client/starwhale/utils/config.py
anda-ren: > ```
>     if platform == "linux" or platform == "linux2":
>         _editor = os.environ.get(SWEnv.sw_editor, "")
>         if not _editor:
>             _editor = os.environ.get("EDITOR", "vi")
>     elif platform == "darwin" or platform == "win32":
>         _editor = os.environ.get(SWEnv.sw_editor, "")
>         if not _editor:
>             raise NoSupportError(f"No {SWEnv.sw_editor} is configured")
>     else:
>         raise NoSupportError(f"Current platform is not supported by: {platform}")
> ```
> 
> -->
> 
> ```
> from shutil import which
> _editor = os.environ.get(SWEnv.sw_editor) or os.environ.get("EDITOR") or "vi"
> if which(_editor) is None:
>    raise noSupportError(f"no found {_editor} bin in {platform}")
> ```
> 
> ?

done

client/starwhale/utils/config.py
anda-ren: > * It would be better if we support `VISUAL` env  https://unix.stackexchange.com/questions/4859/visual-vs-editor-what-s-the-difference
> * `swcli config edit` is not a high-frequency operation, `SWEnv.sw_editor` may not be used by most of the users.
> * `VISUAL` and `EDITOR` should be supported in `darwin`
> * Using a temp file may not be necessary (users my lose "swap detection" features when using vim)
> 
> Refer to the implementation of `pip config edit` https://github.com/pypa/pip/blob/c6783501f16b7ff8c54adf2bcbc1f5f1d11ac699/src/pip/_internal/commands/configuration.py#L225

done

client/starwhale/utils/config.py
anda-ren: tempfile removed as jialei said

client/starwhale/utils/config.py
anda-ren: platform related code is removed

jialeicui (commented):
LGTM

scripts/client_test/cli_test.py
```diff
@@ -101,6 +102,22 @@ def __init__(
...
             logger.info(f"login to server {self.server_url} ...")
             assert self.instance_api.login(url=self.server_url)
+        config.update_swcli_config(
+            **{
+                "link_auths": [
+                    {
+                        "type": "s3",
+                        "ak": "starwhale",
+                        "sk": "starwhale",
+                        "endpoint": "http://10.131.0.1:9000",
```
goldenxinxing, line 112: is there need to use env 'SW_S3_HOST'? the same with "SW_S3_AK","SW_S3_SK",  "SW_S3_EDP", "SW_S3_REGION",

tianweidut (approved):
LGTM

PR #1693, opened 2023-01-03
chore(controller): model serving api refactor

jialeicui:
## Description

- create model serving API returns {id, baseUri} instead of returning id only
- deprecate TTL seconds param 
- allow proxy URI without handler

After this pr, console will do the actions as follow:

1. post creation params
2. try getting the base URI returned in the last step
3. retry when getting 502
4. post user data when getting 200 in the last step

## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


anda-ren (approved):
lgtm

PR #1695, opened 2023-01-04
feat(controller): online eval support gc

jialeicui:
## Description

Online eval support garbage collection
- removes the workloads that reached the max TTL
- removes the oldest workload when there are pending workloads
  - prevent removing the workloads that not reached the min TTL

## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/java/ai/starwhale/mlops/domain/job/ModelServingService.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/domain/job/ModelServingService.java
@@ -43,13 +43,13 @@ import io.kubernetes.client.custom.IntOrString;
 import io.kubernetes.client.openapi.ApiException;
 import io.kubernetes.client.openapi.models.V1ObjectMeta;
 import io.kubernetes.client.openapi.models.V1OwnerReference;
+import io.kubernetes.client.openapi.models.V1Pod;
 import io.kubernetes.client.openapi.models.V1Service;
 import io.kubernetes.client.openapi.models.V1ServicePort;
 import io.kubernetes.client.openapi.models.V1ServiceSpec;
 import io.kubernetes.client.openapi.models.V1StatefulSet;
-import io.kubernetes.client.util.labels.EqualityMatcher;
-import io.kubernetes.client.util.labels.LabelSelector;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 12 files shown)


server/controller/src/main/java/ai/starwhale/mlops/domain/job/ModelServingService.java
```diff
@@ -126,16 +140,23 @@ public ModelServingVo create(
                 .modelVersionId(model.getId())
                 .jobStatus(JobStatus.CREATED)
                 .resourcePool(resourcePool)
+                .lastVisitTime(new Date())
                 .build();
 
-        modelServingMapper.add(entity);
+        long id;
+        synchronized (this) {
+            modelServingMapper.add(entity);
```
goldenxinxing, line 148: Is it more clearly with the name of 'modelServingMapper.insertIgnore(...)'?

server/controller/src/main/java/ai/starwhale/mlops/domain/job/ModelServingService.java
```diff
@@ -126,16 +140,23 @@ public ModelServingVo create(
                 .modelVersionId(model.getId())
                 .jobStatus(JobStatus.CREATED)
                 .resourcePool(resourcePool)
+                .lastVisitTime(new Date())
                 .build();
 
-        modelServingMapper.add(entity);
+        long id;
+        synchronized (this) {
```
goldenxinxing, line 147: Is the synchronized block written to ensure idempotence? if insert is ignore for exception, it seems unnecessary to use sync block

jialeicui: This block is for preventing the race condition with the GC thread, but the synchronized range in the GC thread is wrong, I will change it

server/controller/src/main/java/ai/starwhale/mlops/domain/job/ModelServingService.java
jialeicui: Done in 8f65c60

server/controller/src/main/java/ai/starwhale/mlops/domain/job/ModelServingService.java
jialeicui: Done in 8f65c60

goldenxinxing (approved):
LGTM

PR #1703, opened 2023-01-09
chore(client): make sure model.yaml exists when use custom config file

jialeicui:
## Description

We do not specify the config file name when tasks run in the cluster
- The task run will fail if users do not provide the `model.yaml` config file
- The task will use the wrong config file when users use the custom file and the `model.yaml` exists

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc


goldenxinxing (approved):
LGTM

PR #1707, opened 2023-01-11
fix(e2e): fix console building issue

jialeicui:
## Description

- add `build-all` cmd in console Makefile
- fix gradio building 
- change `yarn build` to `make build-all` in `pub.sh` for e2e building

http://e2e-20230111.pre.intra.starwhale.ai/

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [x] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/scripts/client_test/cli_test.py
+++ b/scripts/client_test/cli_test.py
@@ -273,16 +273,14 @@ class TestCli:
             self.build_runtime(rt["workdir"])
 
         for name, expl in EXAMPLES.items():
-            p = subprocess.Popen(
+            print(f"preparing data for {expl}")
+            rc = subprocess.call(
                 ["make", "CN=1", "prepare-data"],
-                stdout=subprocess.PIPE,
-                stderr=subprocess.STDOUT,
                 cwd=expl["workdir"],
             )
-            with p.stdout:  # type: ignore
-                for line in iter(p.stdout.readline, b""):  # type: ignore
-                    logging.info("got line from subprocess: %r", line)
-            assert not p.wait()
+            if rc != 0:
+                print(f"prepare data for {expl} failed")
+                raise
 
         for name, expl in EXAMPLES.items():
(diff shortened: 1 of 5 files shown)


scripts/client_test/cli_test.py
```diff
@@ -273,16 +273,14 @@ def test_all(self) -> None:
             self.build_runtime(rt["workdir"])
 
         for name, expl in EXAMPLES.items():
-            p = subprocess.Popen(
+            print(f"preparing data for {expl}")
```
goldenxinxing, line 276: use logger.info?

scripts/client_test/cli_test.py
```diff
@@ -273,16 +273,14 @@ def test_all(self) -> None:
...
 
         for name, expl in EXAMPLES.items():
-            p = subprocess.Popen(
+            logger.info(f"preparing data for {expl}")
+            rc = subprocess.call(
                 ["make", "CN=1", "prepare-data"],
-                stdout=subprocess.PIPE,
-                stderr=subprocess.STDOUT,
                 cwd=expl["workdir"],
             )
```
jialeicui, line 280: https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait
> Note This will deadlock when using stdout=PIPE or stderr=PIPE and the child process generates enough output to a pipe such that it blocks waiting for the OS pipe buffer to accept more data. Use [Popen.communicate()](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate) when using pipes to avoid that.

scripts/client_test/cli_test.py
jialeicui: Done

goldenxinxing (approved):
LGTM

PR #1708, opened 2023-01-11
fix(client): datastore loss data on multithread

### LINKED ISSUE #1706: cmp receives support less than ppl output
**Describe the bug**
as shown below, cmp receives 99 output, but ppl process 100
image

**To Reproduce**

**Expected behavior**

**Environment**

goldenxinxing:
## Description
fix #1706 
there will have multi writers on multithreading, so it will lead to race when these writers update the same table. one case is that they all create a table, but only the last would be recorded to the singleton datastore, and then would loss some data which are belongs to the previous table obj.

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/api/_impl/data_store.py
+++ b/client/starwhale/api/_impl/data_store.py
@@ -1085,6 +1085,7 @@ class LocalDataStore:
         self.root_path = root_path
         self.name_pattern = re.compile(r"^[A-Za-z0-9-_/: ]+$")
         self.tables: Dict[str, MemoryTable] = {}
+        self.lock = threading.Lock()
 
     def update_table(
         self,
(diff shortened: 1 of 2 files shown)


client/starwhale/api/_impl/data_store.py
```diff
@@ -1128,6 +1119,23 @@ def update_table(
             else:
                 table.insert(r)
 
+    def _get_table(self, table_name: str, schema: TableSchema) -> MemoryTable:
+        table = self.tables.get(table_name, None)
```
tianweidut, line 1123: remove the duplicated condition check? make all get table operations in the lock context?

client/starwhale/api/_impl/data_store.py
```diff
@@ -1128,6 +1119,23 @@ def update_table(
...
+        if table is not None:
+            return table
+        with LocalDataStore._lock:
```
tianweidut, line 1126: the LocalDataStore _lock is the process scope for all of the objects, but self.tables is the object scope,should we replace the class lock with the object lock?

tianweidut (changes requested):

tianweidut:
@goldenxinxing add some unit tests for the race condition case?

client/starwhale/api/_impl/data_store.py
goldenxinxing: The reason to use double check is to avoid waits for tables which have different names that has already completed initialization,  but i'm not sure that double-checking yields this benefit

client/starwhale/api/_impl/data_store.py
goldenxinxing: done

client/starwhale/api/_impl/data_store.py
goldenxinxing: done

goldenxinxing:
> @goldenxinxing add some unit tests for the race condition case?

done

tianweidut (approved):
LGTM

PR #1727, opened 2023-01-28
fix(client): fix pip cache dir

jialeicui:
## Description

The problem:

```bash
/path/to/venv/pip config get global.cache-dir
/'root'/.cache/pip
```

The single quote `'` makes:
- /'root'/ dir created
- double cache space usage when installing starwhale (if the global starwhale needs to reinstall)
- cache can not be reused between pods

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


goldenxinxing (approved):
LGTM

PR #1755, opened 2023-01-30
chore(example): use the latest starwhale version in pytorch runtime

jialeicui:
## Description

- Rename `requirements-sw-lock` to `requirements.txt` (because it's not the generated version by sw rt lock)
- Removed indirect dependencies (useless and may conflict with the updated pkg, e.g. `boto3==1.21.0` is incompatible with `starwhale>=0.3.6`)

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [x] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


tianweidut (approved):
LGTM

PR #1762, opened 2023-02-01
chore(deploy): remove image pull policy config

jialeicui:
#### Remove the `imagePullPolicy` field

We use the `latest` tag and `IfNotPresent` policy by default, the docker image can not update to the latest version when the image exists in the node the scheduler chooses.
I recommend using k8s default behavior like:

https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting
> When you (or a controller) submit a new Pod to the API server, your cluster sets the imagePullPolicy field when specific conditions are met:
> - if you omit the imagePullPolicy field, and the tag for the container image is :latest, imagePullPolicy is automatically set to Always;
> - if you omit the imagePullPolicy field, and you don't specify the tag for the container image, imagePullPolicy is automatically set to Always;
> - if you omit the imagePullPolicy field, and you specify the tag for the container image that isn't :latest, the imagePullPolicy is automatically set to IfNotPresent.

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [x] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/docker/charts/templates/controller-deployment.yaml
+++ b/docker/charts/templates/controller-deployment.yaml
@@ -31,7 +31,6 @@ spec:
       containers:
         - name: controller
           image: "{{ .Values.image.registry }}/{{ .Values.image.org }}/{{ .Values.image.server.repo }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
-          imagePullPolicy: IfNotPresent
           ports:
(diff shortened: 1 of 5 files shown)


anda-ren (approved):
LGTM

PR #1778, opened 2023-02-06
example(dataset): add kitti lane dataset

anda-ren:
## Description
- add kitti lane dataset
- fix self.resp.raw.read() can't read whole file
image

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/core/dataset/store.py
+++ b/client/starwhale/core/dataset/store.py
@@ -638,7 +638,7 @@ class HttpBufferedFileLike(BaseBufferedFileLike):
             return b""
         elif size < 0:
             # urllib3 resp read does not support to read -1
-            return self.resp.raw.read()  # type: ignore
+            return self.resp.content  # type: ignore
         else:
             return self.resp.raw.read(size)  # type: ignore
 
--- a/client/starwhale/core/dataset/type.py
+++ b/client/starwhale/core/dataset/type.py
@@ -671,9 +671,7 @@ class Text(BaseArtifact, SwObject):
 
     def link_to_content(self, encoding: str = "") -> None:
         if not self.content and self.link:
-            self.content = str(
-                self.link.to_bytes(), encoding or self.encoding
-            )
+            self.content = str(self.link.to_bytes(), encoding or self.encoding)
 
 
 # TODO: support tensorflow transform
--- a/client/starwhale/core/dataset/type.py
+++ b/client/starwhale/core/dataset/type.py
@@ -671,7 +671,10 @@ class Text(BaseArtifact, SwObject):
 
     def link_to_content(self, encoding: str = "") -> None:
         if not self.content and self.link:
-            self.content = str(self.link.to_bytes(self.owner), encoding or self.encoding)
+            self.content = str(
+                self.link.to_bytes(), encoding or self.encoding
+            )
+
 
 # TODO: support tensorflow transform
 # https://cocodataset.org/#format-data
--- a/client/starwhale/core/dataset/type.py
+++ b/client/starwhale/core/dataset/type.py
@@ -646,7 +646,7 @@ class Text(BaseArtifact, SwObject):
         link: t.Optional[Link] = None,
     ) -> None:
         # TODO: add encoding validate
-        self.content = content
+        self._content = content
         super().__init__(
             fp=b"",
             type=ArtifactType.Text,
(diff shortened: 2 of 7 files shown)


client/starwhale/core/dataset/store.py
```diff
@@ -638,7 +638,7 @@ def read(self, size: int) -> bytes:
             return b""
         elif size < 0:
             # urllib3 resp read does not support to read -1
-            return self.resp.raw.read()  # type: ignore
+            return self.resp.content  # type: ignore
```
tianweidut, line 641: `self.resp.content = self.resp.raw.read()` and use the uniform code schema for the different sizes (size <0 and size >= 0).

client/starwhale/core/dataset/type.py
```diff
@@ -659,14 +659,20 @@ def __init__(
...
 
     def to_numpy(self) -> numpy.ndarray:
         return numpy.array(self.to_str(), dtype=self.dtype)
 
-    def to_str(self) -> str:
+    def to_str(self, encoding: str = "") -> str:
+        self.link_to_content(encoding)
         return self.content
 
+    def link_to_content(self, encoding: str = "") -> None:
```
tianweidut, line 672: move these logical into line649 in the __init__ function? so we can get a unique self.content in the whole Text class.

client/starwhale/core/dataset/store.py
anda-ren: sometimes `self.resp.content != self.resp.raw.read()`
```python
>>>import requests
>>> resp=requests.get("https://starwhale-examples.oss-cn-beijing.aliyuncs.com/dataset/kitti/lane/data_road/training/calib/um_000000.txt",stream=True)
>>> b=resp.raw.read()
>>> len(b)
502
>>> resp=requests.get("https://starwhale-examples.oss-cn-beijing.aliyuncs.com/dataset/kitti/lane/data_road/training/calib/um_000000.txt",stream=True)
>>> b=resp.content
>>> len(b)
1860
>>> 

```

client/starwhale/core/dataset/type.py
anda-ren: links are used for lazy initialing. If content in link is fetched in init function, the content will be stored into datastore directlly.

tianweidut: ok.
for the lazy loading, we may use the following schema:

```python
from functools import cached_property

def __init__(self, content):
   self._content = content

@cached_property
def content(self):
   if not self._content:
       self._content = self.link_to_content()

   return self._content

...

call `self.content` property 
````

client/starwhale/core/dataset/store.py
tianweidut: the streaming mode maybe led the unequal value.

anda-ren: > the streaming mode maybe led the unequal value.

Without streaming mode we can't access `resp.raw.read()` like this. See the doc: https://requests.readthedocs.io/en/latest/user/quickstart/#raw-response-content

> In the rare case that you’d like to get the raw socket response from the server, you can access r.raw. If you want to do this, make sure you set stream=True in your initial request. Once you do, you can do this:
> r = requests.get('https://api.github.com/events', stream=True)
> r.raw
> <urllib3.response.HTTPResponse object at 0x101194810>
> r.raw.read(10)
> '\x1f\x8b\x08\x00\x00\x00\x00\x00\x00\x03'

client/starwhale/core/dataset/type.py
anda-ren: > got.
> 
> for the lazy loading, we may use the following schema:
> 
> ```python
> from functools import cached_property
> 
> def __init__(self, content):
>    self._content = content
> 
> @cached_property
> def content(self):
>    if not self._content:
>        self._content = self.link_to_content()
> 
>    return self._content
> 
> ...
> 
> call `self.content` property in `to_bytes` or `to_str` function.
> ```

OK

anda-ren: done

tianweidut (approved):
LGTM

PR #1788, opened 2023-02-06
chore(test): fix dataset test for unexpected .dataset_tmp dir in the code tree

tianweidut:
## Description
image

## Modules
- [x] Client

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/tests/sdk/test_dataset.py
+++ b/client/tests/sdk/test_dataset.py
@@ -7,7 +7,6 @@ import time
 import base64
 import struct
 import typing as t
-import tempfile
 import threading
 from http import HTTPStatus
 from types import TracebackType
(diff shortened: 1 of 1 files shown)


anda-ren (dismissed):
LGTM

client/tests/sdk/test_dataset.py
```diff
@@ -1511,9 +1508,15 @@ def test_row(self) -> None:
 
 
 class TestRowWriter(BaseTestCase):
+    def setUp(self) -> None:
+        super().setUp()
+        self.workdir = Path(self.local_storage) / ".dataset_tmp"
```
jialeicui, line 1513: Is it better to use `os.chdir`? 
And then there is no need to specify the `workdir` param in each test.

tianweidut: updated

jialeicui (approved):
LGTM

PR #1793, opened 2023-02-07
fix(dataset): fix user local fs data can't be fetched by server instance

anda-ren:
## Description
when BaseArtifact.fp is user local file path , it is unreliable and should be removed.
- remove BaseArtifact.fp after being built by dataset if it is user local file path
image

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


tianweidut (approved):
LGTM

PR #1799, opened 2023-02-07
bug(client): fix local wheel package unexpected freezing issue

tianweidut:
## Description
image

By default, the local wheel packages should not be locked into the requirements-sw-lock.txt file which will lead to runtime restore failure.

## Modules
- [x] Client

## Checklist
- [ ] add unit test
- [ ] add necessary doc


jialeicui:
Will the wheel pkg work as expected if it is packaged?

tianweidut:
> Will the wheel pkg work as expected if it is packaged?

yes, the phase of runtime restoring will install whl by the runtime.yaml.  @jialeicui

jialeicui (approved):
LGTM

goldenxinxing (approved):
LGTM

PR #1804, opened 2023-02-08
enhance(cli): make copy command compatible with no `cloud://` prefix uri

tianweidut:
## Description
we cannot execute `swcli runtime cp pytorch-starwhale-main/version/latest pre-k8s/project/mnist ` command without `cloud:// prefix`, following by the user [doc](https://github.com/star-whale/starwhale/blob/main/docs/i18n/zh/docusaurus-plugin-content-docs/current/guides/uri.md#1-instance-uri). 
- before:
  - image
- after:
  - image

## Modules
- [x] Client

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/base/bundle_copy.py
+++ b/client/starwhale/base/bundle_copy.py
@@ -74,8 +74,15 @@ class BundleCopy(CloudRequestMixed):
         else:
             project = None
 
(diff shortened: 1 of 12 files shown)


jialeicui (approved):
LGTM

PR #1807, opened 2023-02-08
feat(datestore): naive version implementation, add records timestamp

xuchuan:
## Modules
- [ ] UI
- [X] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [X] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/java/ai/starwhale/mlops/datastore/DataStore.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/datastore/DataStore.java
@@ -16,7 +16,6 @@
 
 package ai.starwhale.mlops.datastore;
 
-import ai.starwhale.mlops.datastore.MemoryTable.RecordResult;
 import ai.starwhale.mlops.datastore.ParquetConfig.CompressionCodec;
 import ai.starwhale.mlops.datastore.impl.MemoryTableImpl;
 import ai.starwhale.mlops.exception.SwProcessException;
@@ -38,6 +37,7 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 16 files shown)


server/controller/src/main/java/ai/starwhale/mlops/datastore/impl/MemoryTableImpl.java
```diff
@@ -264,19 +293,45 @@ public void update(TableSchemaDesc schema, List<Map<String, Object>> records) {
...
+                deletedFlag = false;
+            }
+            var versions = this.recordMap.computeIfAbsent(key, k -> new ArrayList<>());
+            if (deletedFlag) {
+                if (versions.isEmpty() || !versions.get(versions.size() - 1).isDeleted()) {
+                    versions.add(MemoryRecord.builder()
+                            .timestamp(timestamp)
+                            .deleted(true)
+                            .build());
+                }
```
jialeicui, line 316: Can this block be removed? 
It seems that it's just a special condition of the logic below.

server/controller/src/main/java/ai/starwhale/mlops/datastore/DataStore.java
```diff
@@ -324,27 +353,36 @@ class TableRecords {
...
                     return x.compareTo(y);
                 }).record.getKey();
-                var record = new HashMap<String, Object>();
+                Map<String, Object> record = null;
                 for (var r : records) {
                     if (r.record.getKey().equals(lastKey)) {
-                        record.putAll(
-                                DataStore.encodeRecord(r.meta.columnTypeMap, r.record.values, req.isRawResult()));
+                        if (r.record.isDeleted()) {
+                            record = null;
```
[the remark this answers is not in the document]
xuchuan: "scan" returns a snapshot based on the timestamp parameter. Syncing data between instances with history is another topic. We may add this feature in the future.

server/controller/src/main/java/ai/starwhale/mlops/datastore/impl/MemoryTableImpl.java
xuchuan: Their conditions are different: "!versions.get(versions.size() - 1).isDeleted()" vs "!newRecord.isEmpty()", so it is better to put them in two clauses.

jialeicui (approved):
LGTM

PR #1813, opened 2023-02-09
fix(controller):datastore api return empty results when query multi times for type of bytes

goldenxinxing:
## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc


xuchuan (approved):
LGTM

PR #1817, opened 2023-02-09
feat(controller): populate pypi/docker startup envs to system setting

anda-ren:
## Description
populate pypi/docker startup envs to system setting
- pypi setting in startup env is synced to system setting on startup if user doesn't set them
- system setting have higher priority than startup env
- if user delete them in system setting they will be set to empty and also have higher priority than startup env
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/resources/application.yaml
+++ b/server/controller/src/main/resources/application.yaml
@@ -13,7 +13,7 @@ spring:
 sw:
   version: ${SW_VERSION_CONTROLLER:0.1.0:8c82767b60686f3e2bfea9dafe8c8cce5dd34f52}
   docker:
-    registry: ${SW_DOCKER_REGISTRY_URL:docker-registry.starwhale.cn}
+    registry: ${SW_DOCKER_REGISTRY_URL:}
   runtime:
     pypi:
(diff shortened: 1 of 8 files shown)


server/controller/src/main/resources/application.yaml
```diff
@@ -12,6 +12,8 @@ spring:
       static-locations: file:/opt/starwhale.java/static/, classpath:/static/
 sw:
   version: ${SW_VERSION_CONTROLLER:0.1.0:8c82767b60686f3e2bfea9dafe8c8cce5dd34f52}
+  docker:
+    registry: ${SW_DOCKER_REGISTRY_URL:docker-registry.starwhale.cn}
```
tianweidut, line 16: use empty value as the default?  same as pypi default value.

anda-ren: done

tianweidut (approved):
LGTM

tianweidut:
@anda-ren  update docs?

PR #1893, opened 2023-02-24
feat(console): dataset supports more flexible display, and fix various bugs

waynelwz:
## Description
- The dataset now ensures that different labels have fixed color values under various interaction conditions, and is compatible with both grayscale and colored images.
- Image size adaptation has been added to ensure proper display even when the dataset shape value is incorrect.
- The mask display has been split into different layers to prevent color blending
- Fix evaluation none when request abort caused by route changing

 

## Modules
- [x] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


jialeicui (approved):
LGTM

PR #1934, opened 2023-03-09
chore(client): build support swignore

jialeicui:
## Description

- sw mp build support .swignore
- add ignore config for the present examples

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc


client/starwhale/base/blob/store.py
```diff
@@ -92,7 +92,9 @@ def copy_dir(
         dst_dir = Path(dst_dir)
         excludes = excludes or BuiltinPyExcludes
         for src in src_dir.rglob("*"):
-            exclude = any([fnmatch.fnmatch(str(src), ex) for ex in excludes])
+            exclude = any(
+                [fnmatch.fnmatch(str(src), str(src_dir / ex)) for ex in excludes]
```
tianweidut, line 96: Should `fnmatch` pattern need to bring the path prefix(src_dir)?

jialeicui: Yes

```bash
Python 3.10.9 (main, Dec 19 2022, 17:35:49) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from fnmatch import fnmatch
>>> fnmatch('/foo/bar.txt', 'bar.txt')
False
>>> fnmatch('/foo/bar.txt', '/foo/bar.txt')
True
>>> fnmatch('/foo/baz/bar.txt', 'baz/bar.txt')
False
>>> fnmatch('/foo/baz/bar.txt', '*.txt')
True
>>> fnmatch('/foo/baz/bar.txt', '/foo/*.txt')
True
>>> 
```

tianweidut: got

tianweidut (approved):
LGTM

PR #1937, opened 2023-03-09
chore(client): check if project is set when list bundle

jialeicui:
## Description

Provide users with more friendly tips instead of the error below

image

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


tianweidut (approved):
LGTM

PR #1705, opened 2023-01-10
refactor(dataset): take equal treatment to dataset data and annotation

anda-ren:
## Description
related issue: https://github.com/star-whale/starwhale/issues/1530
implementation:
1. tabular dataset row only contains `idx,data(JsonDict)`
2. only one dataset builder implementation(UserRawBuilder is removed)
3. auto convert bytes/localPath in `data` to uri
4. evaluation related interface change
5. UI related change
6. example related change

## Gains
support below datasets:
- optical flow， 
- scene flow， 
- depth， 
- 3d box， 
- stereo
- pairing
- translation
- Speech synthesis

## Verification
http://row-refactor-20230110.pre.intra.starwhale.ai/

## Modules
- [x] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [x] Python-SDK
- [ ] Others

## Checklist
- [x] run code format and lint check
- [x] add unit test
- [x] add necessary doc

PR #1943, opened 2023-03-13
refactor(dataset): use `features` word to explain data and annotations in the dataset row

tianweidut:
## Description
- The PR only refactors the name and does not change code logic. Replace `DataRow.data` with `DataRow.features`. 
- `features` = `data` + `annotations`. After the https://github.com/star-whale/starwhale/pull/1705 pull request, Starwhale Dataset has a powerful expressiveness for dataset row records. But the `DataRow.data` disambiguates from the user data fields, the example is `row.data.data`, `row.data.label`.
-  The `DataRow.data` is a dict type that is the same as other popular dataset libs that use different words to explain the meaning:
   - [Tensorflow TFDS](https://www.tensorflow.org/datasets/api_docs/python/tfds/features)  -->  `features` 
   - [HuggingFace Datasets](https://huggingface.co/docs/datasets/v2.10.0/en/package_reference/main_classes#datasets.Features) --> `features`
   - [DeepLake](https://docs.deeplake.ai/en/latest/Tensors.html) -> `tensors` 
- Starwale supports `Image`, `Video`, `Audio`, `Text`, `Binary`, and other scalar types in the `features` field. `features` is a more limited word for the purpose. 
- related issue: https://github.com/star-whale/starwhale/issues/1940

## Modules
- [x] Client
- [x] Python-SDK

## Checklist
- [x] add unit test
- [ ] add necessary doc


anda-ren (approved):
LGTM

PR #1966, opened 2023-03-15
fix(client): remove the prefix slash when listing tables in the client

jialeicui:
### Remove the prefix slash when listing tables in the client

Console uses regex `/(^project\/.*?\/.*?\/.*?\/.*?\/)/` to replace the evaluation related info, and the table name with `/` prefix can not match it.

https://github.com/star-whale/starwhale/blob/e1dd759a7803a5879457dc2e272c47de54b1e664/console/packages/starwhale-core/src/widget/utils/replacer.ts#L14-L21

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


tianweidut (approved):
LGTM

PR #1964, opened 2023-03-15
chore(dataset): remove `build_from_handler` interface for dataset python sdk

tianweidut:
## Description
`build_from_handler` may bring confusion for the `Python SDK` style. With Python SDK, users can write in-house code to complete the dataset building and management, the external handler code is not friendly for the users.
In further, `swcli dataset build` command focuses on no-code dataset building. 

related: https://github.com/star-whale/starwhale/issues/1940

## Modules
- [x] Python-SDK

## Checklist
- [x] run code format and lint check
- [x] add unit test
- [ ] add necessary doc

PR #1971, opened 2023-03-15
chore(dataset): new dataset entrypoint function for create or load in Python SDK

tianweidut:
## Description
- dependent pr: https://github.com/star-whale/starwhale/pull/1964
- changes:
  - `create=True` is only for the non-existed dataset.
  - loading an existing dataset, then you can read, update, add, and delete rows or features for the dataset.
  - `readonly`  is an explicit argument for dataset load.
- related issue: https://github.com/star-whale/starwhale/issues/1940

## Modules
- [x] Python-SDK

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/api/_impl/dataset/model.py
+++ b/client/starwhale/api/_impl/dataset/model.py
@@ -84,6 +84,7 @@ class Dataset:
         version: str,
         project_uri: URI,
         create: bool = False,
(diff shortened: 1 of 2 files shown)


anda-ren (approved):
LGTM

PR #1335, opened 2022-10-09
fix(evaluation):add request and limit for step

goldenxinxing:
## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [x] Python-SDK
- [ ] Others

## Checklist
- [ ] run code format and lint check
- [x] add unit test
- [ ] add necessary doc

PR #1975, opened 2023-03-16
fix(controller): jobs from console with GPU requests can not work

jialeicui:
## Description

Add limit for no-k8s resources if not specified by default

related to #1335

## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc


anda-ren (approved):
LGTM

PR #1977, opened 2023-03-16
optimize(console):  avoid rerender and use partical measure columns 

waynelwz:
## Description
1. Optimize the table to decrease the load time from a single 800ms to an average of 250ms and reduce refresh frequency
2. Adjust the calculation logic of the table from all calculations to local recursive computation, only calculated for the displayed content section
3. Restrict the frequent refresh logic caused by column width calculation, complete all asynchronous tasks and limit the trigger frequency to 200ms
Fixes
1. Adjust the view configuration logic to support local storage, prioritizing local configuration > default > all, and resolving the issue of project view not resetting when switching
2. Allow configuration for view to not save global popup prompts
3. Add default content display for grid table
4. Add fill table feature for grid table, with evaluation list not filling by default

## Modules
- [x] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


jialeicui (approved):
LGTM

PR #1993, opened 2023-03-22
refactor(controller): add object storage for dataset

anda-ren:
### AS-IS:
1. BLOB files in dataset are saved in storage under the path of **dataset version**
2. There is no constraint or convention on the name of BLOB files
 
### AFTER THIS PR:
1. BLOB files in dataset are saved in storage under the path of **dataset**
2. The name of BLOB files are the hash of the files by convention

### Why:
The meta info in the same dataset with different versions are save in one datastore table after dataset versioning is done. The file storage of the same dataset with different versions should also has the same storage location in case of tiny change among versions.

### Side effect
4 APIs are added:
1.  POST /project/{projectName}/dataset/{datasetName}/hashedBlob/{hash}
2. HEAD /project/{projectName}/dataset/{datasetName}/hashedBlob/{hash}
3.  GET /project/{projectName}/dataset/{datasetName}/uri
4. POST /project/{projectName}/dataset/{datasetName}/uri/sign-links

2 APIs are deleted:
1. GET /project/{projectUrl}/dataset/{datasetUrl}/version/{versionUrl}/link
2. POST /project/{projectName}/dataset/{datasetName}/version/{version}/sign-links

## Verification
[asciicast](https://asciinema.org/a/LvpfvtimtUxUJANyfMwAMyhkk)
image

## TODO
return absolute path when after saving hashed blob
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


tianweidut (approved):
LGTM

PR #2004, opened 2023-03-23
refactor(client): datastore in standalone support version and mutable schema

jialeicui:
## Description

- The method of data persistence changes from Parquet to JSON
- Support data revision in the scan method
- Table (or row) schema becomes mutable
  - Writing to remote datastore from batch to single, the throughput will decrease
  - Saving the schema in every updating and use the JSON format, the storage space requirements are greatly increasing

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/api/_impl/data_store.py
+++ b/client/starwhale/api/_impl/data_store.py
@@ -1,11 +1,13 @@
+from __future__ import annotations
+
 import os
 import re
 import json
+import time
 import atexit
 import base64
 import struct
 import urllib
-import pathlib
 import binascii
 import importlib
 import threading
@@ -13,7 +15,6 @@ from abc import ABCMeta, abstractmethod
 from http import HTTPStatus
 from typing import (
     Any,
-    Set,
     cast,
     Dict,
     List,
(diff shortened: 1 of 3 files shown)


client/starwhale/api/_impl/data_store.py
```diff
@@ -48,10 +51,12 @@
 except ImportError:
     has_fcntl = False
 
+datastore_table_file_ext = ".swds.json"
```
tianweidut, line 54: `swds` = `starwhale dataset` or `starwhale datastore` ? 

`swds` has already been used for dataset.

tianweidut:
LGTM

client/starwhale/api/_impl/data_store.py
jialeicui: changed to `.sw-datastore.json`

tianweidut (approved):

PR #2014, opened 2023-03-27
fix(e2e): tune e2e script

anda-ren:
## Description
1. thin the size of e2e base image
2. remove thirdparty apt sources after software being installed
3. break e2e script when update system setting  failure occurs
4. break e2e script when apt update failure occurs
5. assume necessary tools are already installed by e2e-base image
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


scripts/e2e_test/start_test.sh
```diff
@@ -274,17 +274,18 @@ install_necessary_tools() {
   if ! command -v unzip &> /dev/null
   then
     echo "installing unzip"
-    apt update && apt install -y unzip
+    apt update
```
goldenxinxing, line 277: if the dockerfile has already install unzip, is there necessary to judge?

anda-ren: > if the dockerfile has already install unzip, is there necessary to judge?

Yes, method `install_necessary_tools` could be removed now

PR #2021, opened 2023-03-27
chore(controller): datastore query and scan api support revision

jialeicui:
## Description

- Use timestamp ms as revision for scan and query
- Ignore the entry with the key only which does not match the revision when scan

## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/java/ai/starwhale/mlops/datastore/impl/MemoryTableImpl.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/datastore/impl/MemoryTableImpl.java
@@ -255,7 +255,7 @@ public class MemoryTableImpl implements MemoryTable {
     }
 
     @Override
-    public String update(TableSchemaDesc schema, @NonNull List<Map<String, Object>> records) {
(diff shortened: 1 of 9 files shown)


server/controller/src/main/java/ai/starwhale/mlops/datastore/impl/MemoryTableImpl.java
```diff
@@ -374,7 +377,10 @@ private Map<String, BaseValue> getRecordMap(BaseValue key, List<MemoryRecord> ve
         if (deleted) {
             ret.put(DELETED_FLAG_COLUMN_NAME, BoolValue.TRUE);
         }
-        ret.put(this.schema.getKeyColumn(), key);
+
+        if (hasVersion) {
+            ret.put(this.schema.getKeyColumn(), key);
+        }
```
jialeicui, line 383: Ignore the key-only entry which does not match the revision @xuchuan

tianweidut (approved):
LGTM

PR #2043, opened 2023-03-30
feat(dataset): add create mode argument for dataset sdk

tianweidut:
- create modes:
  - `auto` mode: If the dataset already exists, creation is ignored. If it does not exist, the dataset is created automatically.
  - `empty` mode: If the dataset already exists, an Exception is raised; If it does not exist, an empty dataset is created. This mode ensures the creation of a new, empty dataset.
  - `forbid` mode: If the dataset already exists, nothing is done.If it does not exist, an Exception is raised. This mode ensures the existence of the dataset.
- create mode can help users to ensure the dataset empty or existence.
- related issue:  #1940 

## Modules
- [x] Python-SDK

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/api/_impl/dataset/model.py
+++ b/client/starwhale/api/_impl/dataset/model.py
@@ -66,6 +66,12 @@ _DEFAULT_LOADER_WORKERS = 2
 _DEFAULT_LOADER_CACHE_SIZE = 20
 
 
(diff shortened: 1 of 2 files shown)


anda-ren (approved):
LGTM

PR #2063, opened 2023-04-08
chore(client): data store support deep copy when scan and insert

jialeicui:
## Description

Data store support deep copy when scan and insert
- insert : always deep copy
- scan: deep copy by default, can be set to false by setting the `SW_DATASTORE_SCAN_DEEP_COPY` env to any string but "true" (case insensitive)

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/api/_impl/data_store.py
+++ b/client/starwhale/api/_impl/data_store.py
@@ -4,6 +4,7 @@ import os
 import re
 import abc
 import sys
+import copy
 import json
 import time
 import atexit
@@ -776,7 +777,9 @@ class InnerRecord:
             self.records = OrderedDict(sorted(self.records.items()))
             self.ordered = True
 
-    def get_record(self, revision: Optional[str] = None) -> Dict[str, Any]:
+    def get_record(
+        self, revision: Optional[str] = None, deep_copy: bool = False
+    ) -> Dict[str, Any]:
         self._reorder()
         ret: Dict[str, Any] = dict()
         for seq, record in self.records.items():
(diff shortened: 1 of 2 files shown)


client/starwhale/api/_impl/data_store.py
```diff
@@ -891,6 +896,7 @@ def scan(
         keep_none: bool = False,
         end_inclusive: bool = False,
         revision: Optional[str] = None,
+        deep_copy: bool = False,
```
tianweidut, line 899: how to use `deep_copy` in the wrapper interface?

jialeicui: The caller chooses whether to copy the data or not because it knows whether it will be modified.
As the doc said ([ref](https://docs.python.org/3/library/copy.html)):
>  A deep copy constructs a new compound object and then, recursively, inserts copies into it of the objects found in the original
> `deep copy` copies everything it may copy too much, such as data which is intended to be shared between copies

tianweidut: for standalone, the scan call chain: dataset wrapper `scan` ->  LocalDataStore `scan_tables` ->  MemoryTable `scan`. 

The `deep_copy` option is only for MemoryTable `scan`.

jialeicui: Made deep copy true by default

tianweidut (approved):
LGTM

PR #2093, opened 2023-04-13
feat(controller): support auto inject user envs to pod

goldenxinxing:
## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/java/ai/starwhale/mlops/domain/settings/SettingsService.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/domain/settings/SettingsService.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2022 Starwhale, Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


server/controller/src/main/java/ai/starwhale/mlops/domain/settings/SettingsService.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public String queryUserSettingsString() {
+        User user = userService.currentUserDetail();
+        var settingsEntity = settingsMapper.get(Scope.USER, user.getId());
+        return settingsEntity == null ? null : settingsEntity.getContent();
+    }
+
+    public Settings queryUserSettings() {
+        try {
+            User user = userService.currentUserDetail();
```
anda-ren, line 52: `userService.currentUserDetail()` relys on http session but there is not http session when controller is deploying tasks. Perhaps this won't work when we call this method in scheduler.

anda-ren (changes requested):

PR #2121, opened 2023-04-20
feat(client): support swcli model info command for output filter

tianweidut:
## Description
- changes:
  - support `--output-filter` option for `swcli model info` command
- command help:
  ```console
  ❯ swcli model info --help
  Usage: swcli model info [OPTIONS] MODEL
  
    Show model details.
  
    MODEL: argument use the `Model URI` format. Version is optional for the
    Model URI. If the version is not specified, the latest version will be used.
  
    Example:
  
            swcli model info mnist # show basic info from the latest version of model
            swcli model info mnist/version/v0 # show basic info from the v0 version of model
            swcli model info mnist/version/latest --output-filter=all # show all info
            swcli model info mnist -of basic # show basic info
            swcli model info mnist -of manifest # show manifest.yaml
            swcli model info mnist -of model_yaml  # show model.yaml
            swcli model info mnist -of handlers # show model runnable handlers info
            swcli model info mnist -of files # show model package files tree
            swcli -o json model info mnist -of all # show all info in json format
  
  Options:
    -of, --output-filter [basic|model_yaml|manifest|handlers|files|all]
                                    Filter the output content. Only standalone
                                    instance supports this option.  [default:
                                    basic]
    --help                          Show this message and exit.
  ```   
- [asciicast](https://asciinema.org/a/578810)

## Modules
- [x] Client

## Checklist
- [x] run code format and lint check
- [x] add unit test
- [x] add necessary doc

PR #2122, opened 2023-04-20
feat(client): support starwhale model package to embed starwhale runtime

tianweidut:
## Description
- depend on: https://github.com/star-whale/starwhale/pull/2121

## Modules
- [x] Client
- [ ] Python-SDK

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/core/model/model.py
+++ b/client/starwhale/core/model/model.py
@@ -51,6 +51,7 @@ from starwhale.base.tag import StandaloneTag
 from starwhale.base.uri import URI
 from starwhale.utils.fs import (
     move_dir,
+    copy_file,
     file_stat,
     ensure_dir,
     ensure_file,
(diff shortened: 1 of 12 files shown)


client/starwhale/core/model/model.py
```diff
@@ -591,8 +613,55 @@ def buildImpl(self, workdir: Path, **kw: t.Any) -> None:  # type: ignore[overrid
...
+
+            info = packaged_runtime
+        else:
+            raise NoSupportError(
+                f"uri type {uri.object.typ} not support to package runtime"
+            )
+
+        dest = self.store.packaged_runtime_bundle_path
+        console.print(f":optical_disk: package runtime({uri}) to {dest}")
+        copy_file(bundle_path, dest)
```
goldenxinxing, line 658: should we ignore the venv/conda which in the origin dir, because copy it to ta new dir will not work in the new path and use many space

goldenxinxing (commented):
others lgtm

client/starwhale/core/model/model.py
tianweidut: bundle_path is a swrt file, which is a tar of runtime

goldenxinxing (approved):
LGTM

PR #2182, opened 2023-05-06
 enhance(sdk): enhance model build sdk for searching modules

tianweidut:
## Description
- changes:
  - Unwrap decorated functions: @pass_context and other decorators can be added for the @handler functions.
  - Ignore the `__main__` module.
  - Tune related-rootdir search modules path for the imported modules scenario.
  - Allow modle build function to be called at the module level, rather than in a cycle call scenario. See the example: https://asciinema.org/a/582707
  - Clarify which modules require explicit arguments and which can be imported implicitly by the model build SDK.
  - Update mnist example and add SDK e2e test.
-  [asciicast](https://asciinema.org/a/582707)

## Modules
- [x] Python-SDK

## Checklist
- [x] add unit test
- [x] add necessary doc

### DIFF
--- a/.github/workflows/client.yaml
+++ b/.github/workflows/client.yaml
@@ -194,4 +194,4 @@ jobs:
         env:
           GITHUB_ACTION: 1
           PYTHON_VERSION: ${{matrix.python-version}}
-        run: bash scripts/client_test/cli_test.sh simple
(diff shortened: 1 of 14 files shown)


anda-ren (approved):
LGTM

PR #2211, opened 2023-05-10
enhance(client): tune model run output for runnable handlers

tianweidut:
## Description
[asciicast](https://asciinema.org/a/583846)

## Modules
- [x] Client

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/base/scheduler/step.py
+++ b/client/starwhale/base/scheduler/step.py
@@ -75,12 +75,20 @@ class Step(ASDictMixin):
         # default run index 0 handler
         job_name = job_name or "0"
         jobs = load_yaml(yaml_path)
-        if job_name in jobs:
-            job = jobs[job_name]
-        else:
-            job_index = int(job_name)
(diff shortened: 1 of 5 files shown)


client/starwhale/base/scheduler/step.py
```diff
@@ -75,12 +75,20 @@ def get_steps_from_yaml(
...
-        if job_name in jobs:
-            job = jobs[job_name]
-        else:
-            job_index = int(job_name)
-            sorted_jobs = sorted(jobs.items())
-            job = sorted_jobs[job_index][1]
+        sorted_jobs = sorted(jobs.items())
+
+        job = None
+        try:
```
anda-ren, line 81: add `Except` clause?  using `try` means there is possible exceptions

tianweidut: Exception or no exception , we can use finally to print the handlers info.

anda-ren (approved):
LGTM

PR #2215, opened 2023-05-10
enhance(client): detect shell conda bin path at first

tianweidut:
## Modules
- [x] Client

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/utils/venv.py
+++ b/client/starwhale/utils/venv.py
@@ -5,6 +5,7 @@ import typing as t
 import platform
 import subprocess
 from pathlib import Path, PurePath, PosixPath
+from functools import lru_cache
 
 import yaml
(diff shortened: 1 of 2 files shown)


client/starwhale/utils/venv.py
```diff
@@ -549,15 +550,18 @@ def get_conda_prefix_path(name: str = "") -> str:
     return output.decode().strip()
 
 
+@lru_cache(maxsize=12)
```
anda-ren, line 553: since the method has no parameter, `maxsize` is needless

tianweidut: ok

anda-ren (approved):
LGTM

PR #2253, opened 2023-05-17
fix(sdk): change the default id format for predict result

goldenxinxing:
## Description
Tune predict result id from "{dataset_name}-{version}__#@#_{ds_index}" to "{dataset_id or name}__#@#_{ds_index}" in cloud and "{project}/{dataset name}__#@#_{ds_index}" in standalone

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [x] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/api/_impl/evaluation.py
+++ b/client/starwhale/api/_impl/evaluation.py
@@ -222,7 +222,13 @@ class PipelineHandler(metaclass=ABCMeta):
             ds.make_distributed_consumption(session_id=self.context.version)
             dataset_info = ds.info
             cnt = 0
-            idx_prefix = _uri.info().get("id") or _uri.name
+            if _uri.instance.is_local:
+                idx_prefix = f"{_uri.project.name}_{_uri.name}"
+            else:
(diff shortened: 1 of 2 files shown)


client/starwhale/api/_impl/evaluation.py
```diff
@@ -222,7 +222,7 @@ def _starwhale_internal_run_predict(self) -> None:
             ds.make_distributed_consumption(session_id=self.context.version)
             dataset_info = ds.info
             cnt = 0
-            idx_prefix = f"{_uri.name}-{_uri.version[:SHORT_VERSION_CNT]}"
+            idx_prefix = _uri.info().get("id") or _uri.name
```
jialeicui, line 225: - raise if `uri.instance.is_cloud` and `uri.info().get("id")` is None?
- use `uri.project.name/uri.name` as prefix if dataset is local ds?

goldenxinxing: done

jialeicui (approved):
LGTM

PR #2255, opened 2023-05-17
fix: canceled job is not deleted

anda-ren:
## Description

Before this pull request, when a `running` task is set to `canceled` the `TaskStatusMachine` considered it illegal. As a result, the logic that deletes the pod wouldn't be executed. This pull request makes this status transfer legal.

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


jialeicui (approved):
LGTM

jialeicui (approved):
LGTM

PR #2261, opened 2023-05-22
chore(client): tune conda log options for the pretty log output

tianweidut:
## Description
current log for conda:
image

use conda `--quiet` to ignore the progress bar output.

## Modules
- [x] Client

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/utils/venv.py
+++ b/client/starwhale/utils/venv.py
@@ -391,7 +391,7 @@ def conda_setup(
     if not name and not prefix:
         raise ParameterError("conda setup must set name or prefix")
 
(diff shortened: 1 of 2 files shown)


client/starwhale/utils/venv.py
```diff
@@ -471,6 +471,8 @@ def conda_env_update(
         get_conda_bin(),
         "env",
         "update",
+        "--verbose",
+        "--quiet",
```
goldenxinxing, line 475: Whether the option of 'quiet' is not effect when 'verbose' is setting.

goldenxinxing (approved):
LGTM

PR #2281, opened 2023-05-26
example(LLM): update example belle-bloom for runtime and dataset ingest

tianweidut:
## Description
- use venv runtime as the default runtime
- handle dataset name pattern

## Modules
- [x] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


anda-ren (approved):
LGTM

PR #2287, opened 2023-05-29
feat(controller): support private resource pool

jialeicui:
## Description

Add two fields to the Resource Pool:
- isPrivate (bool) makes the pool private, and can not be listed when `visibleUserIds` is not specified
- visibleUserIds (array of long), used to configure which users can access this pool

## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


goldenxinxing (approved):
LGTM

PR #2307, opened 2023-06-02
fix(controller): support node select for image build

goldenxinxing:
## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc


server/controller/src/main/java/ai/starwhale/mlops/domain/runtime/RuntimeService.java
```diff
@@ -115,6 +117,7 @@
 @Service
 public class RuntimeService {
 
+    private final String imageBuildResourcePool = "image-build";
```
jialeicui, line 120: Make the pool name configurable?

goldenxinxing: done

jialeicui (approved):
LGTM

PR #2312, opened 2023-06-05
chore(controller): save resource pool detail into step info

jialeicui:
## Description

- Spilt job to step when creating
- Save the resource pool snapshot in step info

## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/java/ai/starwhale/mlops/domain/job/converter/JobBoConverter.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/domain/job/converter/JobBoConverter.java
@@ -45,8 +45,10 @@ import ai.starwhale.mlops.domain.user.bo.User;
 import ai.starwhale.mlops.exception.SwValidationException;
 import ai.starwhale.mlops.exception.SwValidationException.ValidSubject;
 import com.fasterxml.jackson.core.JsonProcessingException;
+import java.io.IOException;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.function.Function;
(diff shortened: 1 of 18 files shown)


server/controller/src/main/java/ai/starwhale/mlops/domain/job/converter/JobBoConverter.java
```diff
@@ -169,19 +171,30 @@ public Job fromEntity(JobEntity jobEntity) {
...
-                            log.error("ERROR!!!!! A job has two running steps job id: {}", job.getId());
-                        }
-                        job.setCurrentStep(step);
-                    }
-                }).collect(Collectors.toList());
+        List<Step> steps = stepEntities.stream().map(entity -> {
+            try {
+                var step = stepConverter.fromEntity(entity);
+                if (step.getResourcePool() == null) {
+                    // backward compatibility
```
anda-ren, line 178: Is there a need for backward compatibility? Not like datasets and models, jobs have very shot life cycles. Could we assume that when we upgrade controller there is no running jobs?

jialeicui: This func is used by `JobBoConverter::fromEntity` and the `fromEntity` func is used in multiple places

goldenxinxing (approved):
LGTM

PR #2350, opened 2023-06-13
enhance(client): add internal model handler for api

jialeicui:
## Description

serving

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/core/model/model.py
+++ b/client/starwhale/core/model/model.py
@@ -248,7 +248,7 @@ class StandaloneModel(Model, LocalStorageBundleMixin):
             show_name="virtual handler for model serving",
             func_name=func.__qualname__,
             module_name=func.__module__,
-            expose=8000,
+            expose=8080,
             virtual=True,
             extra_kwargs={
                 "search_modules": search_modules,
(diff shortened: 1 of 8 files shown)


client/starwhale/core/model/model.py
```diff
@@ -830,6 +850,22 @@ def serve(
...
+        # kwargs see cls._gen_model_serving
+        search_modules = kwargs.get("extra_kwargs", {}).get("search_modules", [])
+        if len(search_modules) == 0:
+            raise ValueError("search_modules is empty")
+
+        workdir = kwargs.get("workdir", "")
+        if not workdir:
+            raise ValueError("workdir is empty")
+
+        port = kwargs.get("expose", 8000)
```
goldenxinxing, line 865: About the expose port, there are two questions:

1. Should we add EXPOSE_PORT to the script of entrypoint?
2. Is Server should parse the spec and add a rule to expose the port?

jialeicui: Like docker's [expose](https://docs.docker.com/engine/reference/builder/#expose)
- The `expose` here does not actually publish the port 
- The `expose` tells the Controller that the job needs to expose a port (the firewall and others should be well configured by the controller, and tell the console the final access URL)
- When getting the expose info, the Controller should proxy the running service (job) if it has no public IP

tianweidut (approved):
LGTM

PR #2370, opened 2023-06-19
enhance(controller): support executing cmd in running task

jialeicui:
## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/java/ai/starwhale/mlops/schedule/k8s/K8sTaskScheduler.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/schedule/k8s/K8sTaskScheduler.java
@@ -44,11 +44,13 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.Future;
 import java.util.stream.Collectors;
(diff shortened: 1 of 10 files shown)


server/controller/src/main/java/ai/starwhale/mlops/schedule/k8s/K8sTaskScheduler.java
```diff
@@ -116,6 +116,24 @@ public void stop(Collection<Task> tasks) {
         });
     }
 
+    @Override
+    public String[] exec(Task task, String... command) {
```
anda-ren, line 120: How about return a Future object , since the `k8sClient.execInPod` is a thread blocking action. Returnning Future object could let people who is using this interface know that the `exec` is a time consuming operation. The tomcat container has thread number limit , abusing this method could cause server panic.

jialeicui: Done

goldenxinxing (approved):
LGTM

PR #2403, opened 2023-06-26
bug(client): skip model list for the empty manifest/digest info

tianweidut:
## Description
image
image

## Modules
- [x] Client

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/core/model/model.py
+++ b/client/starwhale/core/model/model.py
@@ -580,6 +580,9 @@ class StandaloneModel(Model, LocalStorageBundleMixin):
             else:
                 _info = load_yaml(_mpath)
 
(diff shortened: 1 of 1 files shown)


client/starwhale/core/model/model.py
```diff
@@ -580,6 +580,9 @@ def list(
             else:
                 _info = load_yaml(_mpath)
 
+            if not _info:
```
goldenxinxing, line 583: Is there a case that digestFile and manifestFile all of empty in a model package? Is it in an error?

goldenxinxing (approved):
LGTM

PR #2415, opened 2023-06-27
feat(console): use slient config disable error toaster, update model file load api and tree interactive

waynelwz:
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/console/src/api/ApiHeader.tsx
+++ b/console/src/api/ApiHeader.tsx
@@ -76,16 +76,8 @@ function ApiHeader() {
                     return Promise.reject(error)
                 }
 
-                // allow non-exist model file in the eval result page
-                if (
-                    error.response?.status === 404 &&
-                    error.response.config.params?.partName?.includes('eval_panel_layout.json')
-                ) {
-                    return Promise.reject(error)
-                }
-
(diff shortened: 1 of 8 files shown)


console/src/api/ApiHeader.tsx
```diff
@@ -76,16 +76,8 @@ function ApiHeader() {
...
-                    error.response?.status === 404 &&
-                    error.response.config.params?.partName?.includes('eval_panel_layout.json')
-                ) {
-                    return Promise.reject(error)
-                }
-
-                // use user/current as default token auth, it will be triggered multi times, so silent here
-                const withSilentRoute = error.response?.config.url.includes('/user/current')
+                // for example: user/current as default token auth, it will be triggered multi times, so silent here
+                const withSilentRoute = error.response.config.params.silent
```
jialeicui, line 80: `params` will be sent to the server, use hash mark instead?

waynelwz: Server will not handle this parameter, so I think it's will be ok ?

jialeicui (approved):

PR #2457, opened 2023-07-05
chore(controller): prefer using pod failed reason

jialeicui:
## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/java/ai/starwhale/mlops/schedule/k8s/JobEventHandler.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/schedule/k8s/JobEventHandler.java
@@ -149,10 +149,7 @@ public class JobEventHandler implements ResourceEventHandler<V1Job> {
                     taskStatus = TaskStatus.FAIL;
                     stopTime = Util.k8sTimeToMs(condition.getLastTransitionTime());
                     log.debug("job status changed for {} is failed {}", jobName, status);
-                    failedReason = Arrays.stream(new String[] {
-                            condition.getReason(),
-                            condition.getMessage()
-                    }).filter(Objects::nonNull).collect(Collectors.joining(", "));
+                    failedReason = joinReasons(condition.getReason(), condition.getMessage());
                 } else if ("Complete".equalsIgnoreCase(type)) {
                     taskStatus = TaskStatus.SUCCESS;
                     stopTime = Util.k8sTimeToMs(condition.getLastTransitionTime());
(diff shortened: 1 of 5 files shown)


server/controller/src/main/java/ai/starwhale/mlops/schedule/k8s/JobEventHandler.java
```diff
@@ -224,4 +223,18 @@ private Long getPodStartTime(String jobId) {
...
+
+        var failedReasons = pods.stream().filter(p -> p.getStatus() != null)
+                .filter(p -> Objects.equals(p.getStatus().getPhase(), "Failed"))
+                .map(p -> joinReasons(p.getStatus().getReason(), p.getStatus().getMessage()))
+                .filter(Objects::nonNull).collect(Collectors.toList());
+
+        if (failedReasons.size() == 0) {
+            return null;
+        }
+        return failedReasons.stream().min(String::compareTo).get();
```
goldenxinxing, line 238: Is it fit to get reason by the first char of failed reason?

jialeicui: On the native kubelet, the error of the first pod is usually very representative, such as the image does not exist, etc.

If the pod is up, the user can check the cause of the failure directly from the Pod log, which is usually a problem with the user's own program.

server/controller/src/main/java/ai/starwhale/mlops/schedule/k8s/JobEventHandler.java
```diff
@@ -182,6 +179,12 @@ private void updateEvalTask(V1Job job, boolean onDelete) {
             }
         }
 
+        // prefer using the pod failed reason, it has more details
+        var podFailedReason = getPodFailedReason(jobName);
+        if (StringUtils.hasText(podFailedReason)) {
+            failedReason = podFailedReason;
```
anda-ren, line 185: failedReason = f'job failed reason: {failedReason} , pod failed reason: {podFailedReason}' ?

server/controller/src/main/java/ai/starwhale/mlops/schedule/k8s/JobEventHandler.java
anda-ren: If there are more than one pod fail, just concat them?

server/controller/src/main/java/ai/starwhale/mlops/schedule/k8s/JobEventHandler.java
jialeicui: All added

server/controller/src/main/java/ai/starwhale/mlops/schedule/k8s/JobEventHandler.java
jialeicui: All added

anda-ren (approved):
LGTM

PR #2471, opened 2023-07-06
fix(image): use the same logic with client when resolve image's repo

goldenxinxing:
## Description

Replace the built-in image with the repoName(registry/org)

## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc


tianweidut (approved):
LGTM

PR #2497, opened 2023-07-11
feat(controller): build dataset in server

goldenxinxing:
## Description

1. add new operation 'ds_build_and_upload' in entrypoint
2. add dataset build API and support GC the unused files in build record which is succeess
3. add oss signedUrl API
4. add log API(Online and offline)
5. optimise the job and pod eventHandler's dispatch logic(do different operation based on their respective label)

## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc


anda-ren:
maybe we could reuse the job-task framework, so that there is no bother to write log collector/status watcher and token validator

goldenxinxing:
> maybe we could reuse the job-task framework, so that there is no bother to write log collector/status watcher and token validator

Yes, but this need extract some common logics, like whether to sync the job to datastore\ use which resourcePool(image and dataset may use private resource) ..etc. so i will use annother PR(#2509) to fix them.

server/controller/src/main/java/ai/starwhale/mlops/domain/dataset/DatasetService.java
```diff
@@ -371,4 +417,160 @@ public Map<String, String> signLinks(String project, String datasetName, Set<Str
...
+                    request.getDatasetName(), project.getName()));
+            }
+            if (!ds.getDatasetName().equalsIgnoreCase(request.getDatasetName())) {
+                throw new SwValidationException(ValidSubject.DATASET, MessageFormat.format(
+                    "The dataset:{0} is different with the exist dataset:{1}.",
+                    request.getDatasetName(), ds.getDatasetName()));
+            }
+        }
+        var buildings = buildRecordMapper.selectBuildingInOneProjectForUpdate(
+                    project.getId(), request.getDatasetName());
```
jialeicui, line 447: This line means we can apply two buildings with case-sensitive dataset names like `mnist` and `Mnist`?
And it seems that we do not allow this in Line 440:  
```java
if (!ds.getDatasetName().equalsIgnoreCase(request.getDatasetName())) {
```

server/controller/src/main/java/ai/starwhale/mlops/domain/dataset/DatasetService.java
```diff
@@ -371,4 +417,160 @@ public Map<String, String> signLinks(String project, String datasetName, Set<Str
...
+                .projectId(project.getId())
+                .shared(request.getShared())
+                .datasetName(request.getDatasetName())
+                .storagePath(request.getStoragePath())
+                .type(request.getType())
+                .status(BuildStatus.CREATED)
+                .createdTime(new Date())
+                .build();
+        var res = buildRecordMapper.insert(entity) > 0;
+        if (res) {
```
jialeicui, line 464: Is it necessary to check the return value?

server/controller/src/main/java/ai/starwhale/mlops/domain/dataset/DatasetService.java
```diff
@@ -371,4 +417,160 @@ public Map<String, String> signLinks(String project, String datasetName, Set<Str
...
+        var res = buildRecordMapper.updateStatus(id, status) > 0;
+        if (res && status == BuildStatus.SUCCESS && record.getShared()) {
+            // update shared
+            var dataset = datasetMapper.findByName(record.getDatasetName(), record.getProjectId(), false);
+            if (dataset != null) {
+                var version = datasetVersionMapper.findByLatest(dataset.getId());
+                if (version != null) {
+                    datasetVersionMapper.updateShared(version.getId(), true);
+                }
+            }
```
jialeicui, line 544: Make an issue to fix the race condition when the other user updates the dataset?

server/controller/src/main/java/ai/starwhale/mlops/domain/filestorage/FileStorageService.java
```diff
@@ -0,0 +1,145 @@
...
+
+    public FileStorageService(StorageAccessService storageAccessService,
+                              @Value("${sw.filestorage.data-root-path:files}") String dataRootPath,
+                              @Value("${sw.filestorage.url-expiration-time:24h}") String urlExpirationTime) {
+        this.storageAccessService = storageAccessService;
+        if (dataRootPath.endsWith("/")) {
+            this.dataRootPath = dataRootPath;
+        } else {
+            this.dataRootPath = dataRootPath + "/";
+        }
```
jialeicui, line 52: Check if the root path is short enough to make sure that the `generatePathPrefix` less than 255?

server/controller/src/main/java/ai/starwhale/mlops/domain/dataset/DatasetService.java
goldenxinxing: ok, https://github.com/star-whale/starwhale/issues/2523

server/controller/src/main/java/ai/starwhale/mlops/domain/dataset/DatasetService.java
goldenxinxing: yes, you are right, i will fix it

server/controller/src/main/java/ai/starwhale/mlops/domain/filestorage/FileStorageService.java
goldenxinxing: ok

PR #2636, opened 2023-08-21
fix(controller): use lower case name in image build

goldenxinxing:
## Description
fix the error in kaniko when image containers Upper case like 'star-whale/Stable-diffusion-v1-4':
`error checking push permissions -- make sure you entered the correct tag name, and that you are authenticated correctly, and try again: getting tag for destination: repository can only contain the characters abcdefghijklmnopqrstuvwxyz0123456789_-./`

## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/java/ai/starwhale/mlops/domain/runtime/RuntimeService.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/domain/runtime/RuntimeService.java
@@ -696,7 +696,7 @@ public class RuntimeService {
             var user = userService.currentUserDetail();
             var image = new DockerImage(
                     dockerSetting.getRegistryForPull(),
-                    String.format("%s:%s", runtime.getName(), runtimeVersion.getVersionName())
+                    String.format("%s:%s", runtime.getName().toLowerCase(), runtimeVersion.getVersionName())
             );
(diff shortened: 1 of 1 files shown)


server/controller/src/main/java/ai/starwhale/mlops/domain/runtime/RuntimeService.java
```diff
@@ -696,7 +696,7 @@ public BuildImageResult buildImage(String projectUrl, String runtimeUrl, String
             var user = userService.currentUserDetail();
             var image = new DockerImage(
                     dockerSetting.getRegistryForPull(),
-                    String.format("%s:%s", runtime.getName(), runtimeVersion.getVersionName())
+                    String.format("%s:%s", runtime.getName().toLowerCase(), runtimeVersion.getVersionName())
```
jialeicui, line 699: How about adding some comments to let us know the reason for doing the `toLowerCase`

goldenxinxing: done

jialeicui (approved):
LGTM

PR #2683, opened 2023-08-30
fix(controller): add chatset response header

anda-ren:
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


server/controller/src/main/java/ai/starwhale/mlops/api/DataStoreController.java
```diff
@@ -191,6 +192,7 @@ public void scanAndExport(ScanTableRequest request, HttpServletResponse httpResp
                             request.getTables().stream().map(TableDesc::getTableName).collect(
                                     Collectors.toList())) + ".csv\"");
             httpResponse.addHeader("Content-Length", String.valueOf(bytes.length));
+            httpResponse.addHeader("Content-Type", "application/octet-stream;charset=utf-8");
```
jialeicui, line 195: Would it be better to use `Content-Type: text/csv;charset=utf-8`?
https://datatracker.ietf.org/doc/html/rfc4180

jialeicui (approved):
LGTM

PR #2707, opened 2023-09-04
feat(controller): export full table result

anda-ren:
## Description
when limit is smaller than 0 , all results in the table will be exported
cc @waynelwz @lijing-susan 
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/java/ai/starwhale/mlops/api/DataStoreController.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/api/DataStoreController.java
@@ -24,7 +24,6 @@ import ai.starwhale.mlops.api.protocol.datastore.ListTablesRequest;
 import ai.starwhale.mlops.api.protocol.datastore.QueryTableRequest;
 import ai.starwhale.mlops.api.protocol.datastore.RecordListVo;
 import ai.starwhale.mlops.api.protocol.datastore.ScanTableRequest;
-import ai.starwhale.mlops.api.protocol.datastore.TableDesc;
 import ai.starwhale.mlops.api.protocol.datastore.TableNameListVo;
 import ai.starwhale.mlops.api.protocol.datastore.TableQueryFilterDesc;
 import ai.starwhale.mlops.api.protocol.datastore.TableQueryOperandDesc;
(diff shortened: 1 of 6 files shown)


server/controller/src/main/java/ai/starwhale/mlops/api/DataStoreController.java
```diff
@@ -165,14 +165,32 @@ public ResponseEntity<ResponseMessage<RecordListVo>> scanTable(ScanTableRequest
...
+                recordsExporter.exportTo(recordList, outputStream);
+                totalRecords += resultSize;
+                request.setStart(totalRecords);
+                if (remaining != -1) {
+                    remaining = remaining - resultSize;
+                    if (remaining <= 0) {
+                        break;
+                    }
+                }
+                request.setLimit(remaining);
```
jialeicui, line 192: Limit the `limit`?
https://github.com/star-whale/starwhale/blob/12e0fd2887f674e007715ace6a099ae380d53566/server/controller/src/main/java/ai/starwhale/mlops/datastore/DataStore.java#L190-L194

server/controller/src/main/java/ai/starwhale/mlops/api/DataStoreController.java
```diff
@@ -165,14 +165,32 @@ public ResponseEntity<ResponseMessage<RecordListVo>> scanTable(ScanTableRequest
...
+                    "attachment; filename=\"" + extractRealTablename(request.getTableName()) + "."
+                            + recordsExporter.getFileSuffix() + "\""
+            );
             ServletOutputStream outputStream = httpResponse.getOutputStream();
-            outputStream.write(bytes);
+            do {
+                RecordList recordList = queryRecordList(request);
+                int resultSize = recordList.getRecords().size();
+                if (resultSize == 0) {
+                    break;
```
jialeicui, line 181: Would it be better to do like this?
```java
if (resultSize < limit) {
    break;
}
```

server/controller/src/main/java/ai/starwhale/mlops/api/DataStoreController.java
anda-ren: done

server/controller/src/main/java/ai/starwhale/mlops/api/DataStoreController.java
anda-ren: yes, this would reduce one time datastore query

jialeicui (approved):
LGTM

PR #2730, opened 2023-09-09
enhance(client): optimzie the predict count logging for greater accuracy

tianweidut:
## Description

for the following cases:
  - multi datasets
  - `ignore_error=True`, but meets predict exception.
  - batch > 1,  the number of input data is not equal to the number of predict results  

## Modules
- [x] Client

## Checklist
- [ ] add unit test
- [ ] add necessary doc


goldenxinxing (approved):
LGTM

PR #2733, opened 2023-09-11
fix(controller): restore SQL syntax error with multiQueries is true

goldenxinxing:
## Description

There will occur an SQL syntax error when allowMultiQueries=true, the reason is that the content of dump sql script contains semicolon(;) at the line of end.

## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc


anda-ren (approved):
LGTM

PR #2734, opened 2023-09-11
refactor(client): use pydantic model in runtime context

jialeicui:
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/base/cloud.py
+++ b/client/starwhale/base/cloud.py
@@ -296,11 +296,9 @@ class CloudRequestMixed:
         uri_typ: ResourceType,
         page: int = DEFAULT_PAGE_IDX,
         size: int = DEFAULT_PAGE_SIZE,
-        filter_dict: t.Optional[t.Dict[str, t.Any]] = None,
+        filter_dict: t.Optional[t.List[str]] = None,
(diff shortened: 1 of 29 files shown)


client/starwhale/base/cloud.py
```diff
@@ -296,11 +296,9 @@ def _fetch_bundle_all_list(
         uri_typ: ResourceType,
         page: int = DEFAULT_PAGE_IDX,
         size: int = DEFAULT_PAGE_SIZE,
-        filter_dict: t.Optional[t.Dict[str, t.Any]] = None,
+        filter_dict: t.Optional[t.List[str]] = None,
```
tianweidut, line 299: `filter_dict`  is not used in the following function scope,  does this meet expections?

tianweidut:
Others LGTM

client/starwhale/base/cloud.py
jialeicui: All of the API impl do not support filter for now, I will support the filter in the next pr (including the local list filter)

tianweidut (approved):

PR #2746, opened 2023-09-12
fix(controller): reset unprocessed data when job is failed

goldenxinxing:
## Description

1. Remove the back-to-back logic in dataload consumption
2. Add resetUnProcessedData at the task final stage(fail/canceled.)

img_v2_8d48b725-614f-4528-92f1-ea914d021c5g
img_v2_79173a9d-2da3-4c56-a3c8-9e35d9d7026g

## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc


server/controller/src/main/java/ai/starwhale/mlops/domain/job/status/JobUpdateHelper.java
```diff
@@ -90,6 +93,8 @@ public void updateJob(Job job) {
                     TaskStatusChangeWatcher.SKIPPED_WATCHERS.remove();
                 });
             }
+            // deal with the data which belong to current job TODO: the same with retry logic
+            dataLoader.resetUnProcessed(job.getUuid());
```
anda-ren, line 97: Is it better to reset unprocessed data at task level. 
Currentlly job fails if at least one task fails, this logic is not proper in consumption use cases indeed.  The proper logic shall be success at least one success.
Reseting unprocessed data at task level could make us easily correct the status propergation logic.
Moreover, job doesn't necessarily have a dataset consumption.
So better to move this logic to task watcher?

goldenxinxing: done

anda-ren (approved):
LGTM

PR #2760, opened 2023-09-14
feat(controller): controller support rolling update

anda-ren:
## Description

As the Controller is stateful, Controller instance must be shut down for updates.
To minimize the disruption for clients caused by Controller updates, we can keep most of the services who's sates are not maintained by the Controller during the update. We should only pause the state writing service during the transition of state between the old and new machines.

The whole process could be described as the graph below:
image

Controller could start in `rollup` mode or `normal` mode.  Under `rollup` mode, controller waits for old server instance status notifications. While in `normal` mode , controller just start up as if there is no precedence controller instance.

### Copy

Artifacts copy features that use streaming body uploads could not retry by client side when server is switching. This leads to consequences below:
1. Users have to re-upload artifacts mannually in case server is upgrading
3. Dataset build/ runtime dockerizing features in console may fail  at uploading phase when  when server is switching.

### Server switching

1. There must be a reverse proxy on top of server instances to keep consistent endpoint for clients
2. The switching for servers in the reverse proxy is not included in this PR. Users shall do it manually if the reverse proxy doesn't support activate health check.

### DB migration

1. If DB migration fails, users should recover the database using SQL backuped by old instance manually
2. If the old instance is not compitable with new db schema, clients would suffer 500 error for a while

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


server/controller/src/main/java/ai/starwhale/mlops/domain/upgrade/rollup/web/RollingUpdateFilter.java
```diff
@@ -0,0 +1,85 @@
...
+        this.controllerProperties = controllerProperties;
+    }
+
+    @Override
+    protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
+            throws ServletException, IOException {
+        String apiPrefix = controllerProperties.getApiPrefix();
+        if (request.getServletPath().contains(apiPrefix + RollingUpdateController.STATUS_NOTIFY_PATH)
+                || request.getServletPath().contains(apiPrefix + "/login")
+                || request.getServletPath().contains(apiPrefix + "/logout")
```
jialeicui, line 50: Jwt token have no logout?

server/controller/src/main/java/ai/starwhale/mlops/domain/upgrade/rollup/web/UpgradeHealthIndicator.java
```diff
@@ -0,0 +1,75 @@
...
+            primaryInstance = true;
+        }
+    }
+
+    @Override
+    public void onOldInstanceStatus(ServerInstanceStatus status) {
+        details.get(DETAIL_OLD_INSTANCE_STATUSES).add(new Tuple2<>(System.currentTimeMillis(), status));
+        if (status == ServerInstanceStatus.DOWN) {
+            primaryInstance = true;
+        }
```
jialeicui, line 73: When we enable the `roll-up start`
- If the old instance restarts during the updating, it is in an unhealthy status, this may cause the service to be unavailable
- After the update is done, the new instance restart will also cause the service unavailable

It seems that all the related components have the same issue.

docker/charts/templates/controller-deployment.yaml
```diff
@@ -55,26 +55,26 @@ spec:
...
             periodSeconds: 10
             timeoutSeconds: 15
             successThreshold: 1
-            failureThreshold: 5
+            failureThreshold: 60
           startupProbe:
             httpGet:
-              path: /
-              port: {{ .Values.controller.containerPort }}
+              path: /actuator/health
```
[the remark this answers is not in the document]
anda-ren: modified from /actuator/health to /

server/controller/src/main/java/ai/starwhale/mlops/schedule/reporting/SimpleTaskReportReceiver.java
```diff
@@ -33,19 +39,32 @@
...
+        this.taskScheduler = taskScheduler;
+        this.jobBoConverter = jobBoConverter;
     }
 
     @Override
     public void receive(List<ReportedTask> reportedTasks) {
+        if (!primaryInstance) {
+            log.info("server is upgrading and i'm not the primary instance, abandon all reported info");
+            return;
+        }
```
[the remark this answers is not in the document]
anda-ren: alter to Lazy init event watcher

server/controller/src/main/java/ai/starwhale/mlops/domain/upgrade/rollup/web/RollingUpdateFilter.java
anda-ren: but we do have a logout endpoint, leave it here will not effect anything

server/controller/src/main/java/ai/starwhale/mlops/domain/upgrade/rollup/web/UpgradeHealthIndicator.java
anda-ren: rollup start mode configuration moved to database.

jialeicui (approved):
LGTM

PR #2769, opened 2023-09-15
chore(SDK): remove concurrency in sdk/client side

tianweidut:
## Description
- remove `concurreny` in sdk/client side
- fix job/step replicas in the standalone 

## Modules
- [x] Python-SDK

## Checklist
- [ ] add unit test
- [ ] add necessary doc


client/starwhale/base/scheduler/__init__.py
```diff
@@ -42,7 +42,7 @@ def run(
             return self._schedule_all()
 
         if task_index is None or task_index < 0:
-            return [self._schedule_one_step(step_name=step_name, task_num=task_num)]
+            return [self._schedule_one_step(step_name=step_name)]
```
goldenxinxing, line 45: Tasknum in schedule one step function was used to override the stepSpec's 'task_num', Is it should not be deleted?

goldenxinxing (approved):
LGTM

PR #2785, opened 2023-09-22
feat(client): refine project uri in standalone

goldenxinxing:
## Description

**Features:**

1. Unify the used of project unique key(use id in cloud mode and use name in standalone) in http request and file path.
2. Unify the used of project unique key in datastore table name
3. Add log_to_project in model run args so that we can identify the run environment and record environment
4. Pass the full project uri from server

**TODO:**

1. Replace project_get_http_req with ProjectApi.get.

## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc


client/scripts/sw-docker-entrypoint
```diff
@@ -44,6 +44,7 @@ prepare(){
         # only remote
         echo '-->[Preparing] pulling model ...'
         swcli instance login --token "${SW_TOKEN}" --alias server ${SW_INSTANCE_URI}
+        swcli project create ${SW_PROJECT} && swcli project select ${SW_PROJECT}
```
tianweidut, line 47: create `${SW_PROJECT}` project in standalone ?

goldenxinxing: This point is mainly considered when user submit a job to server(such as run in a project named 'demo-a'),but we final submit it to a container, then run it on the local project named 'self' in standalone by default but not 'demo-a'. If user use it to record something, it will be wrong

client/starwhale/base/context.py
```diff
@@ -21,9 +22,12 @@ def __init__(
         index: int = 0,
         dataset_uris: t.Optional[t.List[str]] = None,
         version: str = "",
-        project: str = "",
+        run_project: t.Optional[Project] = None,
+        log_project: t.Optional[Project] = None,
     ):
-        self.project = project
+        self.run_project = run_project or Project()
+        self.project = self.run_project.name
```
tianweidut, line 29: can we replace `self.project` with `self.run_project`?

client/starwhale/core/model/cli.py
```diff
@@ -448,6 +448,13 @@ def _recover(model: str, force: bool) -> None:
     "--run-project",
     envvar=SWEnv.project,
     default="",
+    help=f"Project URI, env is {SWEnv.project}.The model run on the specified project. Default is the current selected project.",
+)
+@optgroup.option(  # type: ignore[no-untyped-call]
+    "-lp",
+    "--log-project",
+    envvar=SWEnv.project,
```
tianweidut, line 456: use another env or remove the env?

client/starwhale/base/uri/project.py
```diff
@@ -40,6 +54,13 @@ def __init__(
             raise Exception("can not init project with empty name")
         self.name = name
         # TODO check if project exists for local and remote
+        if self.instance.is_cloud:
+            # TODO check whether contains namespace in name(like 'sw:project')?
+            self.name = (
+                int(self.name)
+                if self.name.isdigit()
```
tianweidut, line 61: when the type of name is int, `self.name.isdigit()` will raise Exception?

client/starwhale/base/uri/project.py
```diff
@@ -40,6 +54,13 @@ def __init__(
             raise Exception("can not init project with empty name")
         self.name = name
         # TODO check if project exists for local and remote
+        if self.instance.is_cloud:
+            # TODO check whether contains namespace in name(like 'sw:project')?
+            self.name = (
```
tianweidut, line 59: Can we format self.name as string type?

client/scripts/sw-docker-entrypoint
goldenxinxing: removed

client/starwhale/base/context.py
goldenxinxing: done

client/starwhale/core/model/cli.py
goldenxinxing: removed

client/starwhale/base/uri/project.py
goldenxinxing: fixed, unify to use str

client/starwhale/base/uri/project.py
goldenxinxing: done

tianweidut (approved):

PR #2805, opened 2023-09-26
enhance(datastore): allow dumping when updating tables

xuchuan:
## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/java/ai/starwhale/mlops/datastore/DataStore.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/datastore/DataStore.java
@@ -68,7 +68,7 @@ public class DataStore implements OrderedRollingUpdateStatusListener {
     private final StorageAccessService storageAccessService;
 
     private final Map<String, SoftReference<MemoryTable>> tables = new ConcurrentHashMap<>();
-    private final Map<MemoryTable, String> dirtyTables = new ConcurrentHashMap<>();
+    private final Set<MemoryTable> dirtyTables = new HashSet<>();
     private final String snapshotRootPath;
     private final ParquetConfig parquetConfig;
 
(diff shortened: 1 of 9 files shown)


server/controller/src/main/java/ai/starwhale/mlops/datastore/DataStore.java
```diff
@@ -610,13 +622,9 @@ private void clearWalLogFiles() {
...
                 continue;
             }
-            table.lock();
-            try {
-                if (table.getFirstWalLogId() >= 0 && table.getFirstWalLogId() < minWalLogIdToRetain) {
-                    minWalLogIdToRetain = table.getFirstWalLogId();
-                }
-            } finally {
-                table.unlock();
+            var walLogId = table.getFirstWalLogId();
```
jialeicui, line 625: This id may be -1, but the `minWalLogIdToRetain` has changed.

1. mem-table.update.wal-manager.append // wal-manager.max-id changed
2. dump.clear-wal-log.get-wal-max-id and check table.first-wal-id // table.first-wal-id may be -1
3. mem-table.update first-wal-id

server/controller/src/main/java/ai/starwhale/mlops/datastore/impl/MemoryTableImpl.java
```diff
@@ -95,7 +97,7 @@ public class MemoryTableImpl implements MemoryTable {
     private final Map<String, ColumnStatistics> statisticsMap = new HashMap<>();
 
     @Getter
-    private long firstWalLogId = -1;
+    private transient long firstWalLogId = -1;
```
jialeicui, line 100: volatile?

server/controller/src/main/java/ai/starwhale/mlops/datastore/DataStore.java
xuchuan: 1. in this case, the new id will be greater than or equal to the initial minWalLogIdToRetain. Because dumping runs periodically, it is harmless to have a smaller minWalLogIdToRetain.
2. dumping and clearing are always run in the same thread. clearWalLogFiles is always invoked after saveTables. So, this case will never happen.
3. The same as the first case.

server/controller/src/main/java/ai/starwhale/mlops/datastore/impl/MemoryTableImpl.java
xuchuan: Done

jialeicui (approved):
LGTM

PR #2809, opened 2023-09-27
refactor(datastore): use flat structure to avoid too many sub directories

jialeicui:
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/api/_impl/data_store.py
+++ b/client/starwhale/api/_impl/data_store.py
@@ -21,6 +21,7 @@ import threading
 import contextlib
 from abc import ABCMeta, abstractmethod
 from http import HTTPStatus
+from uuid import uuid4
 from typing import (
     Any,
     cast,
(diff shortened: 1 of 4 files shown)


client/starwhale/api/_impl/data_store.py
```diff
@@ -1880,24 +1880,12 @@ def list_tables(
...
-                    datastore_table_file_ext
-                )[0]
-                table_names.add(table_name)
-
-            for table in self.tables:
-                if table.startswith(prefix):
-                    table_names.add(table)
-
-        return list(table_names)
+        manifest = self._load_manifest()
```
tianweidut, line 1883: add 'self.lock' for the `list_tables` function ?

tianweidut:
Others LGTM

client/starwhale/api/_impl/data_store.py
jialeicui: `load_manifest` will see the single version of the content because of the `move` operation of the file,
So I think the lock is useless here.

tianweidut (dismissed):

PR #2819, opened 2023-10-07
enhance(client): make building environment python version as the runtime default python version

tianweidut:
## Modules
- [x] Client

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/core/runtime/model.py
+++ b/client/starwhale/core/runtime/model.py
@@ -1734,7 +1734,11 @@ class StandaloneRuntime(Runtime, LocalStorageBundleMixin):
         runtime_yaml = load_yaml(yaml_path)
 
         mode = runtime_yaml.get("mode", PythonRunEnv.VENV)
-        expected_pyver = str(runtime_yaml.get("environment", {}).get("python", ""))
+        expected_pyver = (
+            str(runtime_yaml.get("environment", {}).get("python", ""))
+            or get_python_version()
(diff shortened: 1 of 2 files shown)


client/starwhale/core/runtime/model.py
```diff
@@ -1963,11 +1967,13 @@ def restore(
...
         _env = _manifest["environment"]
-        _starwhale_version = (
-            _manifest.get("environment", {})
-            .get("lock", {})
-            .get("starwhale_version", "")
+        _lock = _manifest.get("environment", {}).get("lock", {})
+        expected_pyver = (
+            _env["python"]
+            or _lock.get("shell", {}).get("python_version")
+            or get_python_version()
```
jialeicui, line 1974: What is the difference between `environment.python` and `environment.lock.python_version`?
(It seems that we usually use the variable that does not contain the minor version number)

tianweidut: `environment.python` --> defined by the users in the runtime.yaml, the field is optional.
`environment.lock.shell.python_version` --> `swcli runtime build` command uses python bin version that is locked automatically. 

Currently, the 'Starwhale Runtime' only supports Python version formats in the 'major.minor' pattern, such as '3.7' and '3.8'. The 'micro' field will be omitted.

jialeicui (approved):
LGTM

PR #2829, opened 2023-10-09
enhance(dataset): expose exceptions from dataset copy threads

tianweidut:
## Description
- issue:  When exceptions occur within the dataset copy threads, the main thread may fail to detect them.
  [asciicast](https://asciinema.org/a/612774)
- remove some useless `sys.exit`

## Modules
- [x] Client
- [x] Python-SDK

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/core/dataset/copy.py
+++ b/client/starwhale/core/dataset/copy.py
@@ -50,12 +50,7 @@ class DatasetCopy(BundleCopy):
         workers: int = 12,
         **kw: t.Any,
     ) -> None:
-        super().__init__(
-            src_uri,
-            dest_uri,
-            typ=ResourceType.dataset,
-            **kw,
-        )
+        super().__init__(src_uri, dest_uri, typ=ResourceType.dataset, **kw)
 
         self._workers = workers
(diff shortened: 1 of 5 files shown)


client/starwhale/core/dataset/copy.py
```diff
@@ -168,12 +172,13 @@ def _do_dataset_copy(
...
             )
 
+            row_updater_exceptions: t.List[Exception] = []
             row_updater_threads = []
             for i in range(0, self._workers):
                 _t = threading.Thread(
                     name=f"row-updater-{i}",
                     target=self._do_row_update,
-                    args=(progress, task_id, dest),
+                    args=(progress, task_id, dest, row_updater_exceptions),
```
goldenxinxing, line 181: Are there has a way to fail fast, such as to define a callback function which deal with the exception, like this:
`def exception_callback(e: Exception) -> None:`
`  raise e`
so that the main thread can fail fast as soon as.

tianweidut: Python does not support thread callback yet.

goldenxinxing (approved):
LGTM

PR #2837, opened 2023-10-10
bug(sdk): fix list/tuple types check for converting sequence

tianweidut:
## Description
test for `example/datasets/LLM/KQApro`
image

## Modules
- [x] Python-SDK

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/core/dataset/tabular.py
+++ b/client/starwhale/core/dataset/tabular.py
@@ -13,7 +13,12 @@ from collections import UserDict, defaultdict
 
 from typing_extensions import Protocol
 
-from starwhale.utils import console, gen_uniq_version, validate_obj_name
+from starwhale.utils import (
+    console,
+    gen_uniq_version,
+    validate_obj_name,
+    check_python_scalar_type,
+)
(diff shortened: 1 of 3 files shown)


client/starwhale/core/dataset/tabular.py
```diff
@@ -213,10 +218,15 @@ def _transform(data: t.Any) -> t.Any:
...
-                else:
+                if (
+                    not types
+                    or len(types) == 1
+                    and check_python_scalar_type(types.pop())
+                ):
+                    # Only keep the simple format(the scalar type in list/tuple) the original type for dataset viewer, such as: [1,2,3]
                     return data
+                else:
+                    return Sequence(data=data, auto_convert=True)
```
jialeicui, line 229: With this logic, the `[[1, 2], [3, 4]]` will use `Sequence`, is this situation as expected?

tianweidut: yes

jialeicui (approved):
LGTM

PR #2838, opened 2023-10-10
enhance(datastore): support multiple type of list

jialeicui:
## Description

image

## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/api/_impl/data_store.py
+++ b/client/starwhale/api/_impl/data_store.py
@@ -131,36 +131,32 @@ class SwType(metaclass=ABCMeta):
     @staticmethod
     def decode_schema(schema: Dict[str, Any]) -> "SwType":
         type_name = schema.get("type", None)
+
+        def decode_list_types(
+            _schema: Dict[str, Any]
+        ) -> Tuple[SwType, Dict[int, SwType] | None]:
+            element_type = _schema.get("elementType", None)
+            if element_type is None:
+                raise RuntimeError(f"no element type found for type {type_name}")
+            _attrs = _schema.get("attributes", None)
+            sparse_types: Dict[int, SwType] | None = None
+            if _attrs is not None:
+                if not isinstance(_attrs, list):
+                    raise RuntimeError("attributes should be a list")
+                sparse_types = {}
+                for _attr in _attrs:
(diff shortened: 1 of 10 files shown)


client/starwhale/api/_impl/data_store.py
```diff
@@ -127,12 +137,30 @@ def decode_schema(schema: Dict[str, Any]) -> "SwType":
...
-                raise RuntimeError(f"no element type found for type {type_name}")
-            return SwTupleType(SwType.decode_schema(element_type))
+            element_types = schema.get("elementTypes", None)
+            if element_types is None:
+                element_type = schema.get("elementType", None)
+                if element_type is None:
+                    raise RuntimeError(f"no element type found for type {type_name}")
+                element_types = [element_type]
+            decoded_element_types = [SwType.decode_schema(t) for t in element_types]
+            return SwTupleType(decoded_element_types)
```
goldenxinxing, line 163: Should there get the attrs and pass it to SwTupleType constructor?

client/starwhale/api/_impl/data_store.py
```diff
@@ -386,151 +429,168 @@ def serialize(self, value: Any) -> Any:
...
+                return x[1] > y[1] and 1 or -1
 
+            sorted_types = sorted(
+                types.items(), reverse=True, key=cmp_to_key(count_then_type)
+            )
+            most_common_type = sorted_types[0][0]
+            self.main_type = most_common_type
+            assert isinstance(element_types, list)
+            for i, typ in enumerate(element_types):
+                if typ != most_common_type:
```
goldenxinxing, line 480: Would this lead to this error: original list is [1,'2','3', 4.0], but when i want to get the type for index of 2(str), then i find the float type(because the main type is str, and sparse_types is [int, float])?

client/starwhale/api/_impl/data_store.py
jialeicui: Done

anda-ren (approved):
LGTM

PR #2844, opened 2023-10-11
chore(client): fix mypy lint issue

jialeicui:
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/core/instance/view.py
+++ b/client/starwhale/core/instance/view.py
@@ -59,7 +59,7 @@ class InstanceTermView(BaseTermView, CloudRequestMixed):
             sys.exit(1)
 
     def info(self, uri: str = "") -> t.Dict:
-        instance = Instance(uri or self.current_instance)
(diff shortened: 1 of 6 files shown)


client/starwhale/core/instance/view.py
```diff
@@ -59,7 +59,7 @@ def logout(self, instance: str) -> None:
             sys.exit(1)
 
     def info(self, uri: str = "") -> t.Dict:
-        instance = Instance(uri or self.current_instance)
+        instance = Instance(uri)
```
goldenxinxing, line 62: Will this change lead to breaking change, because the Instance constructor don't deal with the uri=""?

goldenxinxing (approved):
LGTM

PR #2846, opened 2023-10-11
chore(controller): add max thread property for server and tune default to 500

goldenxinxing:
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/resources/application.yaml
+++ b/server/controller/src/main/resources/application.yaml
@@ -158,6 +158,9 @@ mybatis:
   mapper-locations: classpath*:mapper/**/*.xml
 
 server:
+  tomcat:
+    threads:
+      max: ${SW_CONTROLLER_MAX_THREADS:500}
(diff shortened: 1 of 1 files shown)


server/controller/src/main/resources/application.yaml
```diff
@@ -158,6 +158,9 @@ mybatis:
   mapper-locations: classpath*:mapper/**/*.xml
 
 server:
+  tomcat:
+    threads:
+      max: ${SW_CONTROLLER_MAX_THREADS:500}
```
jialeicui, line 163: Does this number include the built-in thread pool?

goldenxinxing: yes, renamed it

PR #2864, opened 2023-10-16
fix(controller): fix jobs that run on k8s are always cancelling when cancel operation is commited

anda-ren:
## Description
fix jobs that run on k8s are always cancelling when cancel operation is commited
Tests have passed in the cases:
1. When pod is pending
2. When container is creating
3. When container is running
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/java/ai/starwhale/mlops/schedule/impl/k8s/reporting/PodEventHandler.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/schedule/impl/k8s/reporting/PodEventHandler.java
@@ -73,11 +73,6 @@ public class PodEventHandler implements ResourceEventHandler<V1Pod> {
             return;
         }
 
-        // https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination
-        if (pod.getMetadata() != null && pod.getMetadata().getDeletionTimestamp() != null) {
-            log.info("pod {} is being deleted", pod.getMetadata().getName());
-            return;
-        }
         Long rid = getJobNameAsId(pod);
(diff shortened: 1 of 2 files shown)


server/controller/src/main/java/ai/starwhale/mlops/schedule/impl/k8s/reporting/PodEventHandler.java
```diff
@@ -73,19 +73,18 @@ private void reportRunStatus(V1Pod pod) {
...
             return;
         }
 
         RunStatus runStatus;
         var phase = pod.getStatus().getPhase();
-        if (StringUtils.hasText(phase)) {
+        // https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination
+        if (pod.getMetadata() != null && pod.getMetadata().getDeletionTimestamp() != null) {
+            log.info("pod {} is being deleted", pod.getMetadata().getName());
+            runStatus = RunStatus.FAILED;
```
jialeicui, line 86: Will this operation turn a successful run into a failed one?

anda-ren: > Will this operation turn a successful run into a failed one?

No, it is not blocked by other modules. 
The logic here is not accurate, though. And, I have modified it.

jialeicui (approved):
LGTM

PR #2865, opened 2023-10-16
fix(controller): fix when explicit_defaults_for_timestamp is off, run table could not be created

anda-ren:
fix when explicit_defaults_for_timestamp is off, run table could not be created

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


jialeicui (approved):
lgtm

PR #2873, opened 2023-10-17
enhance(datastore): map support multiple key value schemas

jialeicui:
## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/pyproject.toml
+++ b/client/pyproject.toml
@@ -18,8 +18,8 @@ exclude = '''
     | example/LLM/belle-bloom/BELLE
     | example/LLM/llama2/llama
     | starwhale/proto_gen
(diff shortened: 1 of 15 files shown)


anda-ren:
The sparse map with index as the key may not work properlly.

See quote from https://datatracker.ietf.org/doc/html/rfc8259.html
> An object is an unordered collection of zero or more name/value
   pairs, where a name is a string and a value is a string, number,
   boolean, null, object, or array.
>   An array is an ordered sequence of zero or more values.

tianweidut (approved):

PR #2895, opened 2023-10-23
chore(controller): datastore use read-write lock

goldenxinxing:
## Description
Test case(use 200 thread to scan the same table which has 20w data)
image

Speed up the scan processing on multi scan request case.

## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


jialeicui (approved):
LGTM

PR #2913, opened 2023-10-27
chore(client): prefer using builtin list when building dataset

jialeicui:
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/core/dataset/tabular.py
+++ b/client/starwhale/core/dataset/tabular.py
@@ -210,6 +210,8 @@ class TabularDatasetRow(ASDictMixin):
                 return Binary(fp=data, auto_convert_to_bytes=True)
             elif isinstance(data, dict):
                 return {k: _transform(v) for k, v in data.items()}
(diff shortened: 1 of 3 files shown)


client/starwhale/core/dataset/tabular.py
```diff
@@ -215,18 +210,6 @@ def _transform(data: t.Any) -> t.Any:
                 return Binary(fp=data, auto_convert_to_bytes=True)
             elif isinstance(data, dict):
                 return {k: _transform(v) for k, v in data.items()}
-            elif isinstance(data, (list, tuple)):
```
tianweidut, line 218: Only remove `Sequence`, keep `transform` for list and tuple?

jialeicui: Fixed

tianweidut (approved):

PR #2927, opened 2023-10-31
enhance(client): prefer using the built-in map when building dataset

jialeicui:
## Description

Dataset support dict with int key now.

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


tianweidut (approved):
LGTM

PR #2945, opened 2023-11-08
enhance(SDK): enhance hf datasets to handle abnormal dataset without config class

tianweidut:
## Description
- issue:
   - hf raw data:   https://huggingface.co/datasets/lambdalabs/pokemon-blip-captions
     image
   - hf lib:
     VRECFKrW41
   - starwhale dataset: 
    image
 - fixed:
   image

## Modules
- [x] Client
- [x] Python-SDK

## Checklist
- [x] run code format and lint check
- [x] add unit test
- [ ] add necessary doc

PR #2947, opened 2023-11-08
chore(sdk): revert "enhance(SDK): enhance hf datasets to handle abnormal dataset without config class"

tianweidut:
Reverts star-whale/starwhale#2945

image

When datasets >= `2.14.5`, the datasets can return the expected features, so we ignore workaround.


goldenxinxing (approved):
LGTM

PR #2969, opened 2023-11-13
fix(client): fix broken resource ls cmd

jialeicui:
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


client/starwhale/base/uri/project.py
```diff
@@ -112,15 +112,12 @@ def __eq__(self, other: object) -> bool:
...
+def get_remote_project_id(instance_uri: str, token: str, project: str) -> Any:
     resp = requests.get(
         urllib.parse.urljoin(instance_uri, f"/api/v1/project/{project}"),
         headers={
             "Content-Type": "application/json; charset=utf-8",
-            "Authorization": (
-                SWCliConfigMixed().get_sw_token(instance=instance_uri)
-                or os.getenv(SWEnv.instance_token, "")
-            ),
+            "Authorization": token,
```
jialeicui, line 120: Automatically getting token using uri can not work properly when multiple instances aliases for the url are used.

tianweidut (approved):

PR #2991, opened 2023-11-17
enhance(client): support gradio components in api decorator

jialeicui:
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/api/_impl/service/types.py
+++ b/client/starwhale/api/_impl/service/types.py
@@ -1,4 +1,10 @@
 from enum import Enum
+from typing import Any
+
+try:
+    import gradio
+except ImportError:
+    gradio = None
 
 
 class ServiceType(Enum):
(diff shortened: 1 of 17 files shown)


client/starwhale/api/_impl/service/types.py
```diff
@@ -1,4 +1,10 @@
 from enum import Enum
+from typing import Any
+
+try:
+    import gradio
```
tianweidut, line 5: move `import gradio` into the function `all_components_are_gradio`?

tianweidut:
Others LGTM

client/starwhale/api/_impl/service/types.py
jialeicui: Done

tianweidut (approved):

PR #2999, opened 2023-11-20
refactor(storage): remove autofit package

xuchuan:
## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc


anda-ren:
Could you provide a concise list of the benefits that this refactoring brings? Upon reviewing this pull request, I noticed that instead of adding code, we should modify the existing code to introduce a new storage implementation. This modification is violating the open-closed principle.

server/storage-access-layer/src/main/java/ai/starwhale/mlops/storage/StorageAccessService.java
```diff
@@ -16,15 +16,107 @@
...
+        return storageAccessServiceMap.computeIfAbsent("memory", key -> new StorageAccessServiceMemory());
+    }
+
+    static StorageAccessService getFileStorageAccessService(FsConfig config) {
+        return storageAccessServiceMap.computeIfAbsent("file-" + config.hashCode(),
+                key -> new StorageAccessServiceFile(config));
+    }
+
+    static StorageAccessService getS3LikeStorageAccessService(String type, S3Config config) {
+        return storageAccessServiceMap.computeIfAbsent(type.toLowerCase() + "-" + config.hashCode(), key -> {
```
jialeicui, line 102: hash code is too easy to [collision](https://classes.engineering.wustl.edu/2019/spring/cse247/studios/9/)
Assuming that only the regions in the config are inconsistent, then whether the final result of the hash code is consistent depends on whether the regions will collide.

```java
public void testStringHashCodeCollision() {
    var hashTable = new HashMap<Integer, String>();
    // generate all string hash code with two char string
    for (int i = 'A'; i < 'z'; i++) {
        for (int j = 'A'; j < 'z'; j++) {
            var str = new String(new byte[] { (byte) i, (byte) j });
            var hashCode = str.hashCode();
            if (hashTable.containsKey(hashCode)) {
                System.out.println("hash code collision: " + str + " and " + hashTable.get(hashCode));
            }
            hashTable.put(hashCode, str);
        }
    }
}
```

part of the outputs

```
hash code collision: BB and Aa
hash code collision: BC and Ab
hash code collision: BD and Ac
hash code collision: BE and Ad
hash code collision: BF and Ae
hash code collision: BG and Af
hash code collision: BH and Ag
hash code collision: BI and Ah
hash code collision: BJ and Ai
hash code collision: BK and Aj
hash code collision: BL and Ak
hash code collision: BM and Al
hash code collision: BN and Am
hash code collision: BO and An
hash code collision: BP and Ao
hash code collision: BQ and Ap
hash code collision: BR and Aq
hash code collision: BS and Ar
hash code collision: BT and As
```

xuchuan:
> Could you provide a concise list of the benefits that this refactoring brings? Upon reviewing this pull request, I noticed that instead of adding classes, we should modify the existing class code to introduce a new storage implementation. This modification is violating the open-closed principle. @xuchuan

The primary purpose of this refactoring is to make adding a new S3-like implementation much easier. In the old way, we had to modify four classes: the implementation itself, the autofit adaptation, the configuration, and the blob service. These classes are in different packages, and it is very easy to omit some of them. After this modification, only add an implementation and register it in StorageService.
The second one is to unify the configuration storage type and URI scheme. In origin, they are hardcoded in different places.  Now, S3-like storage types and schemes are all registered in StorageService.
The third one is to make the code cleaner. The whole autofit package is removed.  We deducted 800 lines of code.

server/storage-access-layer/src/main/java/ai/starwhale/mlops/storage/StorageAccessService.java
xuchuan: Done. Change the key to a pair.

jialeicui:
LGTM

anda-ren (approved):
LGTM

PR #3023, opened 2023-11-23
fix(controller): fix user input duplicate env error when finetuning

anda-ren:
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/java/ai/starwhale/mlops/domain/ft/FineTuneAppService.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/domain/ft/FineTuneAppService.java
@@ -65,6 +65,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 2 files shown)


server/controller/src/main/java/ai/starwhale/mlops/domain/ft/FineTuneAppService.java
```diff
@@ -241,7 +243,10 @@ private String rewriteSpecEnvForRequest(
                 }
                 Map<String, String> envMap = env.stream()
                         .filter(e -> null != e.getValue())
-                        .collect(Collectors.toMap(Env::getName, Env::getValue));
+                        .collect(Collectors.toMap(Env::getName, Env::getValue, (env1, env2) -> {
+                            log.warn("duplicate env detected from user, random value {} is chosen", env1);
```
goldenxinxing, line 247: Is it more suitable with the logic of coverage if we use the last one

goldenxinxing (approved):

PR #3027, opened 2023-11-24
fix(client): make llm chat service works with pydantic v2

jialeicui:
## Description

- FastAPI uses return type annotation to generate router, the typing system does not support `|`
- Pydantic v2 optional field should specify `= None` if the construct function does not always contain the field.
- Top K to int

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


tianweidut (approved):
LGTM

PR #3029, opened 2023-11-24
feat(controller): add domain names for object storage

anda-ren:
## Description

recognize regex pattern for the object store domain in the header

client part:
https://github.com/star-whale/starwhale/pull/3048

## Test cases
image

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc


server/controller/src/main/java/ai/starwhale/mlops/configuration/security/ObjectStoreDomainDetectionFilter.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+        Map<String, String> endpointEquivalentsMap = s3Config.getEndpointEquivalentsMap();
+        domainAliasMap = endpointEquivalentsMap.entrySet().stream().collect(Collectors.toMap(Entry::getKey, (entry) -> {
+            URI uri = null;
+            try {
+                uri = new URI(entry.getValue());
+            } catch (URISyntaxException e) {
+                throw new RuntimeException(e);
+            }
+            return Pattern.compile(uri.getHost().replace(".", "\\."));
```
jialeicui, line 59: What if the host configured by the user is the same, but the sub-path is different? Can we directly let the OSS impl class support sign by name?

anda-ren: > What if the host configured by the user is the same, but the sub-path is different? Can we directly let the OSS impl class support sign by name?

If the host is the same while the sub-path is different, they are not equivalent

anda-ren: > What if the host configured by the user is the same, but the sub-path is different? Can we directly let the OSS impl class support sign by name?

If the OSS impl class support sign by name , every impl class has to do similar implementation which is code duplication indeed. And that breaks the original interface which will cause clients' code change

jialeicui (approved):
LGTM

PR #3041, opened 2023-11-28
feat(controller): add docker dispatched container to a network

anda-ren:
## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/java/ai/starwhale/mlops/schedule/impl/docker/RunExecutorDockerImpl.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/schedule/impl/docker/RunExecutorDockerImpl.java
@@ -59,8 +59,6 @@ public class RunExecutorDockerImpl implements RunExecutor {
     final String nodeIp;
     final String networkName;
 
-    Network network;
-
     static final String NETWORK_HOST = "host";
 
     public RunExecutorDockerImpl(
--- a/server/controller/src/main/java/ai/starwhale/mlops/schedule/impl/docker/RunExecutorDockerImpl.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/schedule/impl/docker/RunExecutorDockerImpl.java
@@ -33,7 +33,6 @@ import com.github.dockerjava.api.exception.DockerException;
 import com.github.dockerjava.api.model.Container;
 import com.github.dockerjava.api.model.Frame;
 import com.github.dockerjava.api.model.HostConfig;
-import com.github.dockerjava.api.model.Network;
 import com.github.dockerjava.api.model.PullResponseItem;
 import java.io.Closeable;
 import java.io.IOException;
--- a/server/controller/src/main/java/ai/starwhale/mlops/schedule/impl/docker/RunExecutorDockerImpl.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/schedule/impl/docker/RunExecutorDockerImpl.java
@@ -18,6 +18,8 @@ package ai.starwhale.mlops.schedule.impl.docker;
 
 import ai.starwhale.mlops.domain.run.bo.Run;
 import ai.starwhale.mlops.domain.run.bo.RunStatus;
+import ai.starwhale.mlops.exception.SwValidationException;
+import ai.starwhale.mlops.exception.SwValidationException.ValidSubject;
 import ai.starwhale.mlops.schedule.executor.RunExecutor;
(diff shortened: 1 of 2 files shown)


tianweidut (approved):
LGTM

server/controller/src/main/java/ai/starwhale/mlops/schedule/impl/docker/RunExecutorDockerImpl.java
```diff
@@ -131,10 +150,14 @@ public void onComplete() {
                     createContainerCmd.withCmd(containerCommand.getCmd());
                 }
                 CreateContainerResponse createContainerResponse = createContainerCmd.exec();
+                dockerClient.connectToNetworkCmd()
+                        .withNetworkId(networkName)
+                        .withContainerId(createContainerResponse.getId())
+                        .exec();
```
jialeicui, line 156: How does the controller connect to the container through this network?

server/controller/src/main/java/ai/starwhale/mlops/schedule/impl/docker/RunExecutorDockerImpl.java
```diff
@@ -55,28 +57,44 @@ public class RunExecutorDockerImpl implements RunExecutor {
...
+                if (null == network) {
+                    try {
+                        network = dockerClient.inspectNetworkCmd().withNetworkId(this.networkName).exec();
+                    } catch (NotFoundException e) {
+                        dockerClient.createNetworkCmd().withName(this.networkName).exec();
+                        network = dockerClient.inspectNetworkCmd().withNetworkId(this.networkName).exec();
+                    }
+                }
+            }
+        }
```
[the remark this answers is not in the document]
anda-ren: > https://codeql.github.com/codeql-query-help/java/java-unsafe-double-checked-locking/
> 
> Just ensure the network in the constructor?

- First let's talk about the article you quote. This article argues that double checked locking is not thread safe because of returning partially initiallized object or memory inconsistent object. It gives three examples, two of which returns  partially initiallized object . The last example returns inconsitent object. But this won't happen in our case. Quote from [oracle](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/package-summary.html#MemoryVisibility):

> Each action in a thread happens-before every action in that thread that comes later in the program's order.

This rule guarentees that Line94 happens before Line95, which means `network` is fetched after the network creating action successfully returned.
https://github.com/star-whale/starwhale/blob/525e311d953b318dc0ffba62916846d4f465c9f1/server/controller/src/main/java/ai/starwhale/mlops/schedule/impl/docker/RunExecutorDockerImpl.java#L94-L95

> An unlock (synchronized block or method exit) of a monitor happens-before every subsequent lock (synchronized block or method entry) of that same monitor. And because the happens-before relation is transitive, all actions of a thread prior to unlocking happen-before all actions subsequent to any thread locking that monitor.

This rule guarentees that the initialization of `network` happens before all subsequent threads that reads `network`
https://github.com/star-whale/starwhale/blob/525e311d953b318dc0ffba62916846d4f465c9f1/server/controller/src/main/java/ai/starwhale/mlops/schedule/impl/docker/RunExecutorDockerImpl.java#L88-L99

- Second , why not put it in the constructor. Because the 'dockerClient' can only be got in `run` method within current framework. Change the framework is another issue :) 
https://github.com/star-whale/starwhale/blob/525e311d953b318dc0ffba62916846d4f465c9f1/server/controller/src/main/java/ai/starwhale/mlops/schedule/impl/docker/RunExecutorDockerImpl.java#L84-L86

server/controller/src/main/java/ai/starwhale/mlops/schedule/impl/docker/RunExecutorDockerImpl.java
anda-ren: `host` network which is the default behavor is kept to be supported after commit [keep support for host network](https://github.com/star-whale/starwhale/pull/3041/commits/99c76a63a69545fc89c24bc7a8675e62469ee514) .
If user choose to use network other than `host` he/she should guarentee that controller is in the same network as did by https://github.com/tianweidut/starwhale/blob/a6dd437b92d91bcf6a4d375b82643361feb702c3/client/starwhale/cli/server/compose.tmpl

jialeicui (approved):
LGTM

PR #3082, opened 2023-12-14
feat(client): support `swcli runtime run` command

tianweidut:
## Description
`swcli runtime activate` + `run command` + `deactivate` three commands -->  `swcli runtime run` 

```bash
❯ swcli runtime run --help
Usage: swcli runtime run [OPTIONS] URI

  Run a command in the Starwhale Runtime environment.

  For an existed Starwhale runtime, the command will restore and activate the
  runtime automatically.

  Positional arguments:

          URI: Starwhale Runtime URI in the standalone instance
          CMD: Executable command name, with additional arguments if needed.

  Examples:

          # run some python code in the starwhale runtime
          swcli runtime run pytorch python -c "import torch;print(torch.__version__)"

          # run python script in the starwhale runtime with verbose mode
          swcli -vvv runtime run --cwd example/helloworld helloworld python3 dataset.py

          # use the interactive Python shell
          swcli runtime run pytorch --live-stream ipython3

Options:
  --cwd TEXT     Working directory
  --live-stream  Do not capture stdout/stderr, when the is an interactive cmd,
                 like ipython, the option should be enabled
  --help         Show this message and exit.
```

[asciicast](https://asciinema.org/a/626817)

## Modules
- [x] Client

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/core/runtime/process.py
+++ b/client/starwhale/core/runtime/process.py
@@ -47,10 +47,11 @@ class Process:
     def __repr__(self) -> str:
         return f"process: {self._mode} in prefix path {self._prefix_path}, runtime uri: {self._uri}"
 
-    def run(self) -> None:
-        if os.environ.get(self.EnvInActivatedProcess, "0") == "1":
-            raise RuntimeError("already in runtime activated process")
+    def run_self_swcli_cmd(self) -> None:
+        """Run the original swcli command in the runtime environment.
 
+        The function will remove `--runtime` option and replace the proper swcli bin path.
+        """
         argv = copy.deepcopy(sys.argv)
         if len(argv) < 2:
             raise NoSupportError("no cli command")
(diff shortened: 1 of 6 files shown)


client/starwhale/core/runtime/process.py
```diff
@@ -72,19 +73,36 @@ def run(self) -> None:
...
+            cmd = " ".join([f"'{c}'" for c in cmd])
 
         # TODO: support windows platform
         if self._mode == PythonRunEnv.VENV:
             _bin_path = (self._prefix_path / "bin/activate").absolute()
-            cmd = " && ".join([f"source {_bin_path}", sub_cmd])
+            cmd = " && ".join([f"source {_bin_path}", cmd])
         elif self._mode == PythonRunEnv.CONDA:
-            cmd = f"{get_conda_bin()} run --live-stream --prefix {self._prefix_path.absolute()} {sub_cmd}"
+            cmd = f"{get_conda_bin()} run --live-stream --prefix {self._prefix_path.absolute()} {cmd}"
```
goldenxinxing, line 98: are there should judge the param?
```
if live_stream:
  cmd = f"{get_conda_bin()} run --live-stream --prefix {self._prefix_path.absolute()} {cmd}" 
elif
  cmd = f"{get_conda_bin()} run --prefix {self._prefix_path.absolute()} {cmd}"
```

tianweidut: no need. `--live-stream` is only used for starwhale popen function.

goldenxinxing (approved):
LGTM

PR #3083, opened 2023-12-15
fix(client): the online evaluation is accessible behind the proxy sub-path

jialeicui:
## Description

demo: http://jialei.pre.intra.starwhale.ai/projects/1/jobs/8/servings

Bug reason:  the front end can not access the assets and the APIs if the server is behind the proxy sub-path

I have implemented a simple string replacement on the client side to achieve the effects of injecting global variables and modifying asset paths. 
I did not use templates like Jinja2 because once templates are used, the client/index.html file can no longer be used for debugging and packaging with Vite.

## Modules
- [ ] UI
- [ ] Controller
- [ ] Agent
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/api/_impl/service/service.py
+++ b/client/starwhale/api/_impl/service/service.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import os
 import sys
+import json
 import typing as t
 import functools
 
@@ -14,7 +15,7 @@ else:
 
 from fastapi import FastAPI
 from pydantic import Field
-from starlette.responses import FileResponse
+from starlette.responses import HTMLResponse
 from starlette.staticfiles import StaticFiles
 
 from starwhale.utils import console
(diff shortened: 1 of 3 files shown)


client/starwhale/api/_impl/service/service.py
```diff
@@ -149,8 +150,21 @@ def spec() -> t.Union[ServiceSpec, None]:
...
+            root = os.getenv("SW_ONLINE_SERVING_ROOT_PATH", "")
+            opt = {
+                "root": root,
+            }
+            script = f"""<script type="text/javascript">
+                window.starwhaleConfig = {json.dumps(opt)}
+                </script>"""
+            content = content.replace("<!-- starwhale-client-placeholder -->", script)
+            # replace assets path
+            content = content.replace("/assets/", f"{root}/assets/")
```
anda-ren, line 166: If user access to the service by ip directlly this still makes resources points to proxy backed urls? Can we add a header flag to make the api server aware of wether it is behind a proxy or not.

PR #3086, opened 2023-12-15
enhance(client): support to detect shell activated runtime uri for model build

tianweidut:
## Modules
- [x] Client

## Checklist
- [ ] add unit test
- [ ] add necessary doc


client/starwhale/utils/venv.py
```diff
@@ -685,24 +688,34 @@ def activate_python_env(mode: str, identity: str, interactive: bool) -> None:
         if not _bin.startswith("/") or _name == _bin:
             _bin = shutil.which(_name) or _bin
 
+        envs = os.environ.copy()
+        envs[SWShellActivatedRuntimeEnv.MODE] = mode
+        envs[SWShellActivatedRuntimeEnv.PREFIX] = identity
+        envs[SWShellActivatedRuntimeEnv.URI] = original_runtime_uri
```
goldenxinxing, line 694: Is it possible to unify the envs with the two cases that RuntimeProcess.run(which used for model run in a new subProcess) and StandaloneRuntime.activate(which used for activate a shell to replace current process)?

tianweidut: Using the different envs have some benfits:
- use case: `swcli runtime activate`，then run `swcli model run xxx --runtime`.  The subprocess env has the higher priority.
- To clarify the runtime uri from subprocess(command scope) or shell(shell scope).

goldenxinxing: ok

goldenxinxing (approved):
LGTM

PR #3092, opened 2023-12-20
chore(controller): add run info to pod env

goldenxinxing:
## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/server/controller/src/main/java/ai/starwhale/mlops/schedule/SwTaskSchedulerImpl.java
+++ b/server/controller/src/main/java/ai/starwhale/mlops/schedule/SwTaskSchedulerImpl.java
@@ -30,6 +30,7 @@ import ai.starwhale.mlops.schedule.impl.container.ContainerSpecification;
 import ai.starwhale.mlops.schedule.impl.container.TaskContainerSpecificationFinder;
 import ai.starwhale.mlops.schedule.reporting.RunReportReceiver;
 import java.text.MessageFormat;
+import java.util.HashMap;
 import java.util.concurrent.Future;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.context.annotation.Lazy;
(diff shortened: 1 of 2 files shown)


server/controller/src/main/java/ai/starwhale/mlops/schedule/SwTaskSchedulerImpl.java
```diff
@@ -67,27 +67,30 @@ public void schedule(Task task) {
...
             runMapper.insert(runEntity);
             Run run = Run.builder()
                     .id(runEntity.getId())
                     .taskId(task.getId())
-                    .runSpec(runSpec)
                     .logDir(runEntity.getLogDir())
                     .build();
             task.setCurrentRun(run);
+
+            ContainerSpecification containerSpecification = taskContainerSpecificationFinder.findCs(task);
```
jialeicui, line 82: Would it be better if the function's parameter was set as Run?
There would be no need to ensure that task's `currentRun` has been assigned value beforehand. Otherwise, there's an implicit constraint where setCurrentRun must be called before using this function.

server/controller/src/main/java/ai/starwhale/mlops/schedule/SwTaskSchedulerImpl.java
```diff
@@ -67,27 +67,30 @@ public void schedule(Task task) {
...
+            RunSpec runSpec = RunSpec.builder()
+                    .image(containerSpecification.getImage())
+                    .envs(containerSpecification.getContainerEnvs())
+                    .command(containerSpecification.getCmd())
+                    .resourcePool(task.getStep().getResourcePool())
+                    .requestedResources(task.getTaskRequest().getRuntimeResources())
+                    .build();
+
+            runMapper.updateSpec(runEntity.getId(), runSpec);
```
anda-ren, line 91: just modify the two lines below is enough, I think.

https://github.com/star-whale/starwhale/blob/02fd459fac2491777889fccb054e2f79bbe2deeb/server/controller/src/main/java/ai/starwhale/mlops/schedule/impl/k8s/RunExecutorK8s.java#L74
https://github.com/star-whale/starwhale/blob/02fd459fac2491777889fccb054e2f79bbe2deeb/server/controller/src/main/java/ai/starwhale/mlops/schedule/impl/docker/RunExecutorDockerImpl.java#L133

server/controller/src/main/java/ai/starwhale/mlops/schedule/SwTaskSchedulerImpl.java
anda-ren: The initializing of run depends on `containerSpecification`, so the parameter is task.

anda-ren (approved):
LGTM

tianweidut (approved):
LGTM

PR #3117, opened 2024-01-09
feat: add custom args

goldenxinxing:
## Description

1. write arguments spec to job.yaml
2. support parse arguments in server side
3. output the failed log in githua action

## Modules
- [ ] UI
- [x] Controller
- [x] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/api/_impl/argument.py
+++ b/client/starwhale/api/_impl/argument.py
@@ -14,7 +14,7 @@ from typing_extensions import Literal
 from starwhale.utils import console
 from starwhale.utils.pydantic import PYDANTIC_V2
 from starwhale.base.models.model import OptionFieldClient
-from starwhale.base.client.models.models import OptionField
+from starwhale.base.client.models.models import Arguments
 
 
 # TODO: use a more elegant way to pass extra cli args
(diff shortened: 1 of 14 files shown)


client/starwhale/api/_impl/argument.py
```diff
@@ -94,14 +65,25 @@ def asdict(self) -> t.Dict[str, t.Any]:
...
-                    field = OptionField(**option.to_info_dict())
+                    field = OptionFieldClient(**option.to_info_dict())
                     if PYDANTIC_V2:
                         info = field.model_dump(mode="json")
                     else:
                         info = field.dict()
                     r[func][dtype][option.name] = info
         return r
 
+    def asobj(self) -> t.Dict[str, t.Dict[str, t.Dict[str, OptionField]]]:
```
tianweidut, line 76: define a type alias for `t.Dict[str, t.Dict[str, OptionField]]`?  StepSpec.arguments field also needs this type.

client/starwhale/base/client/models/models.py
```diff
@@ -1480,6 +1480,26 @@ class ServiceSpec(SwBaseModel):
...
+    name: str
+    param_type: str
+    # for Choice type
+    choices: Optional[List[str]] = None
+    case_sensitive: bool = False
+
+
+class OptionField(SwBaseModel):
+    name: str
+    opts: List[str]
```
tianweidut, line 1493: Maybe we can remove `opts` field, only use `name` field as input argument name.

tianweidut:
Others LGTM

client/starwhale/base/client/models/models.py
goldenxinxing: removed

client/starwhale/api/_impl/argument.py
goldenxinxing: done，but StepSpec.arguments was generated by pydantic, so just used in argument.py

tianweidut (approved):

PR #3128, opened 2024-01-16
example(LLM): change chatglm3 generation_config field type for command line arguments

tianweidut:
## Modules
- [x] Others

## Checklist
- [ ] add unit test
- [ ] add necessary doc

### DIFF
--- a/example/llm-finetune/models/chatglm3/finetune.py
+++ b/example/llm-finetune/models/chatglm3/finetune.py
@@ -72,6 +72,10 @@ class TrainArguments(Seq2SeqTrainingArguments):
     remove_unused_columns: bool = dataclasses.field(
         default=False, metadata={"help": "remove unused columns"}
     )
+    # Currently, we don't support `Optional[Union[str, Path, GenerationConfig]]` for the command line.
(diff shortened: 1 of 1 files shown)


example/llm-finetune/models/chatglm3/finetune.py
```diff
@@ -72,6 +72,10 @@ class TrainArguments(Seq2SeqTrainingArguments):
     remove_unused_columns: bool = dataclasses.field(
         default=False, metadata={"help": "remove unused columns"}
     )
+    # Currently, we don't support `Optional[Union[str, Path, GenerationConfig]]` for the command line.
+    generation_config: t.Optional[str] = dataclasses.field(
```
goldenxinxing, line 76: Does this property is unused?

tianweidut: we need to change the default value for the command line. The HF trainer will use the property.

goldenxinxing: ok，i see

goldenxinxing (approved):

PR #3131, opened 2024-01-17
fix(controller): support multiple value

goldenxinxing:
## Description
when the argument is multiple, like this:
image

the expect action is:
image

## Modules
- [ ] UI
- [x] Controller
- [ ] Agent
- [ ] Client
- [ ] Python-SDK
- [ ] Others

## Checklist
- [x] add unit test
- [ ] add necessary doc

### DIFF
--- a/client/starwhale/base/client/models/models.py
+++ b/client/starwhale/base/client/models/models.py
@@ -1392,7 +1392,7 @@ class OptionField(SwBaseModel):
     type: OptionType
     required: bool
     multiple: bool
-    value: Optional[str] = None
+    value: Optional[Dict[str, Any]] = None
     help: Optional[str] = None
     hidden: bool
     default: Optional[str] = None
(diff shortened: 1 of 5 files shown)


client/starwhale/base/client/models/models.py
```diff
@@ -1392,7 +1392,7 @@ class OptionField(SwBaseModel):
     type: OptionType
     required: bool
     multiple: bool
-    value: Optional[str] = None
+    value: Optional[Dict[str, Any]] = None
```
jialeicui, line 1395: It seems that the server will parse it as a bool or list. Would there be any issues using a dict in Python?

goldenxinxing: The value is unused in client now.
What's more uncomfortable is that java's obj is converted into dict when passing through pydantic.

